PR #1, opened 2018-03-28
Drawing automata with networkx library and count num words accepted of specific length

guiem:
Nice library.

I added a function to draw the automata, you can check how it looks like here: https://github.com/guiem/my_notebooks/blob/master/automata/dfa.ipynb

Also added a function to compute the number of words accepted by the automata of a specific length (it comes handy to solve certain automata problems).

Because of the level of inheritance I don't know if I placed the code in the right classes. If not, and you don't have time to do it, just tell me where the code should go and I will place it correctly.

I'm aware this adds dependency on more libraries, I proposed the pull request just in case it was of your interest, otherwise I will keep developing the fork. Thanks again for such a nice work implementing this library!

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -3,6 +3,7 @@
 
 import copy
 import queue
+import itertools
 
 import automata.fa.fa as fa
 import automata.shared.exceptions as exceptions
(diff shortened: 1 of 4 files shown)


automata/fa/dfa.py
```diff
@@ -162,3 +163,12 @@ def _init_from_nfa(self, nfa):
             states=dfa_states, input_symbols=dfa_symbols,
             transitions=dfa_transitions, initial_state=dfa_initial_state,
             final_states=dfa_final_states)
+
+    def num_accepted(self, length):
```
caleb531: This function would be better served returning a generator instead of the full list. In practice, that means you would `yield w` within your loop instead of building up the `accepted` list of results. In this case, you would still have access to the list of accepted permutations, and if you need the length, you could `len(list(dfa.num_accepted(length=3)))`

guiem:
Hi @caleb531 !

I'm closing it myself. I totally understand your reasons, and I honestly doubted you could (or would like to) merge it as it is. As I told you it was "just in case".

I'll keep an eye on v2 and meanwhile I'll patch some of my specific needs. And btw, I really really appreciate your comments and recommendations, I found all of them really useful. **I truly appreciate that you took the time to give me that beautiful feedback**!

caleb531:
@guiem I just realized Automata v2 if you were interested to see what's changed:
https://github.com/caleb531/automata/releases/tag/v2.0.0

PR #2, opened 2018-05-02
Add __init__.py to every module.

YtvwlD:
Those files not existing trips up some interpreters (eg. bpython).


caleb531:
@YtvwlD Good call—thanks for the PR. :)

rjptechsolutions:
Can you give some proper examples of it how to implement this Lib... Like nfa to dfa convert  with user input

YtvwlD:
There is a [README](https://github.com/caleb531/automata#dfafrom_nfacls-nfa) which should answer this question.

caleb531:
@rjptechsolutions In particular, see the `DFA.from_nfa()` section (and the section immediately following, which shows you how to create an NFA):
https://github.com/caleb531/automata#dfafrom_nfacls-nfa

PR #3, opened 2018-05-03
Optimize and improve DFA.from_nfa (and add a test for this change).

YtvwlD:
I had an issue converting a NFA (which is a little bit complex) to a DFA, so I took a look at how the conversion works. I came up with these changes:

 * There's no need to re-compute transitions we've already done, so let's skip those.
 * Also let's rely on the queue to determine whether we're done.
   This seems to produce more correct results.

I tried to add a minimal example to the tests.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -143,11 +143,13 @@ class DFA(fa.FA):
 
         state_queue = queue.Queue()
         state_queue.put({nfa.initial_state})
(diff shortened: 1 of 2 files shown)


caleb531:
@YtvwlD This is great! Thanks for the contribution—I had thought my original implementation of `DFA.from_nfa()` would already cover all cases, but apparently not.

YtvwlD:
Thanks for merging. :)
I'm really glad that this library even exists.

PR #4, opened 2018-05-18
Minify DFAs

YtvwlD:
A given DFA can be minified by first removing unreachable states and then merging equivalent states.

I've implemented `DFA.minify` (which seems to work, but I'm not exactly sure), added two tests for this (I didn't add tests for `_remove_unreachable_states`, `_compute_reachable_states`, `_create_markable_states_table`, `_mark_states_table_first`, `_mark_states_table_second` and `_join_non_marked_states`) and added a bit of documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,15 @@ dfa.validate()  # returns True
 dfa.copy()  # returns deep copy of dfa
 ```
 
+#### DFA.minify(self)
+
+Creates a minimal DFA which accepts the same inputs as the old one.
+Unreachable states are removed and equivalent states are merged.
+
+```python
+minimal_dfa = dfa.minify()
+```
+
 #### DFA.from_nfa(cls, nfa)
 
 Creates a DFA that is equivalent to the given NFA.
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -195,11 +195,11 @@ class DFA(fa.FA):
                     continue
                 # merge them!
                 s3 = s.union(s2)
-                # add the new one
-                non_marked_states.add(s3)
                 # remove the old ones
                 non_marked_states.remove(s)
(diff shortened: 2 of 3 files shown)


caleb531:
@YtvwlD Thank you for your contribution—this is very well done. You seemed to pick up on and follow all the implicit project conventions (docstrings, tests, even the README docs).

No worries about not having a test for every method—I tend to only write tests for public API methods (i.e. those not prefixed with `_`) *or* internal API methods that are generic enough to be used more than once.

However, I have two changes to request before I'm ready to merge this in:

## Remove `assert` statements

Please remove your `assert` statements for `dfa.validate()`; They add extra computations, additional side effects (mid-computation). The `assert` statements seem more useful for debugging and verifying the logical integrity of the implementation, but I am convinced that unit tests would be better for achieving that goal (and more consistent with the rest of the project). In other words, there are no `assert` statements; there are only tests. ;)

Besides, I'd personally prefer `DFA.validate()` to remain a method only invoked on instantiation or manually by the user.

## Fix missing code coverage

Per the `coveralls` bot above, there are a few lines of code in your PR which are not covered in your unit tests ([see the report here](https://coveralls.io/builds/17060062/source?filename=automata%2Ffa%2Fdfa.py)). I ask that you please add additional tests to bring that coverage back up to a glorious 100%.

Thanks,
Caleb

YtvwlD:
Thank you for your feedback.

I've changed the `isinstance` condition and removed the `assert` statements.

I'll do something about the test coverage soon.

YtvwlD:
So, this I think this will do.

Sorry for the delay, I really thought everything was ready when I created this PR.

caleb531:
@YtvwlD Beautiful! Everything is looking just right. Merging in now...

PR #7, opened 2018-06-12
TM: Support not moving (direction N)

YtvwlD:
Today, I first tried simulating a DTM. And it didn't work, because the DTM I tried had transitions which didn't move the head.

This PR quickly implements this feature.

A few thoughts:
 * I didn't change the DTM in the README to make use of this.
 * I don't know if this `elif direction == 'N': pass` makes sense.
   But because the other two cases were present in the function, I figured it would be nice.
 * I didn't add any tests. (There's just a little bit new code, if any. - Let's see if Coveralls complains.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,8 @@ dtm = DTM(
 )
 ```
 
+The direction `N` (for no movement) is also supported.
+
 #### DTM.read_input(self, input_str, step=False)
 
 Returns a tuple containing the final state the machine stopped on, as well as a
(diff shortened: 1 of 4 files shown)


caleb531:
@YtvwlD Neat—I never knew Turing machines supported an `N` direction. I welcome this addition.

The `not in` check and the existing `# pragma: no branch` keep Coveralls from complaining, but I'm not letting you off the hook. :P One of the purposes of automated testing is to show that you need the code you wrote to achieve the desired behavior (in your case, marking `N` as a valid direction).

Therefore, I ask that you add a test for a DTM with an `N` transition. You don't need to bother adding a `README` example, though.

P.S. I agree about the `elif` for `N`—keep it.

YtvwlD:
You're absolutely right; I should've added tests.

I added a new DTM with those transitions.

caleb531:
@YtvwlD Beautiful! Let me merge this in.

PR #10, opened 2018-08-26
Add support for NTMs

YtvwlD:
This pull request adds a new class `NTM` which represents nondeterministic Turing machines. It is conceptually very similar to #9.

This might break code which relies on the string representations of the tape or configuration of Turing machines. Otherwise, it should be fine, I think.

So, what it does is basically the following:
 * `TMTape` becomes immutable and hashable. It still represents the tape of a TM and the current cursor position. (I've removed the offset stuff; I'm not sure about this.)
 * `TMConfiguration` is new. This is a (immutable and hashable) class representing the current configuration of a TM. It is basically a tuple of state and tape. Also, `tmtools.print_config` has been moved here.
 * `DTM` has been adjusted to these changes.
 * `DTM._get_next_configuration` is the main thing, now.
 * `DTM._has_accepted` is new. (It does what the name says it does.)
 * `NTM` is new. It is very similar to `DTM`; the main difference is that it holds a set of current configurations instead of just one.
 * Tests and documentation for all of this have been added.

One thing that came to my mind during this is that the `copy` method on now-immutable types (such as `TMTape`) should probably be removed, because they have no effect and might cause confusion.
But this also affects #9 - so this probably warrants a new pull request?

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,11 +584,19 @@ The direction `N` (for no movement) is also supported.
 
 #### DTM.read_input(self, input_str)
 
-Returns a tuple containing the final state the machine stopped on, as well as a
+Returns a `TMConfiguration`. This is basically a tuple containing the final state the machine stopped on, as well as a
 `TMTape` object representing the DTM's internal tape (if the input is accepted).
 
 ```python
-dtm.read_input('01')  # returns ('q4', TMTape('xy.'))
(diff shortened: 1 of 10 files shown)


caleb531:
@YtvwlD Once again, absolutely marvelous job. I'll merge it in now.

And yes, I agree `copy()` can be removed for the immutable types, but as you suspected, I would definitely prefer that as a separate PR.

Issue #8, opened 2018-08-07
Conditions of PDA Acceptance


YtvwlD:
DPDAs can accept when the stack is empty or a final state has been reached (assuming that all input has been read).

The related code is `DPDA._check_for_input_rejection` which is basically this:
```python
if current_state not in self.final_states and stack:
            raise exceptions.RejectionException()
```

This leads to the problem that when the DPDA accepts via final state, it will also accept if the stack is empty after reading a given input, even when no final state has been reached.

I would prefer specifying whether a DPDA accepts either via empty stack or via final state.

(This can probably be deducted by looking at whether `self.final_states` is empty but I'm not so sure about that.)

caleb531:
@YtvwlD Any update on how you'd like to handle PDA acceptance? I'd like to get Automata v3 released soon, but would also like to hear your input first in case any more breaking API changes need to be made.

YtvwlD:
I'm really sorry to not have responded earlier. I really hope to get to this next week.

caleb531:
It's quite alright, Niklas! :) I look forward to hearing your response when you can get to it.

Caleb

caleb531:
Hey, @YtvwlD! I just wanted to check in if you happened to have some time this week to comment more on this issue. If not, that's alright—I'm just checking. :)

YtvwlD:
No, sorry. But I hope to be able to take a look before the year ends.

caleb531:
@YtvwlD That's fine! :)

caleb531:
@YtvwlD I'm eager to start writing release notes and get Automata v3 published since more people are finding this library (no doubt thanks to your efforts in making it more comprehensively useful). I am still interested in hearing your feedback, but will require now that any changes you recommend be backwards-compatible with the pending v3 API (because I am not interested in pushing a v4 for the sake of this one behavioral change).

That being said, I have done some research on the topic myself and have concluded that there's no hard rule mandating that a given PDA cannot accept by both modes simultaneously. Besides, if I recall correctly, one can simply add a lambda/epsilon transition at the start of an automaton to avoid the automatic empty-state acceptance.

Ultimately, I'd still consider an additional parameter for the user to customize the behavior on-the-fly, but I am definitely committed to keeping the current behavior as the default. I will keep this issue open for the sake of discussion and because this is still under consideration.

YtvwlD:
Sorry for not getting to this earlier.

[...]

PR #12, opened 2019-01-31
Make acceptance mode of PDAs configurable.

YtvwlD:
This is related to issue #8.

I implemented the feature (which was a small change) and adjusted the documentation and tests.
I didn't add a test for the case where one mode is specified and the other occurs, though.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -25,8 +25,8 @@ but `tuple` does not.
 ## Acceptance mode of PDAs is now configurable
 
 `DPDA` and `NPDA` have a new config option which specifies when to accept.
-This has to be either `empty_stack`, `final_state` or `both`.
-The latter was the default in v2.
+This can be either `empty_stack`, `final_state` or `both`.
+The latter was the default in v2 and still is.
 
 ## Backwards-incompatible changes from v1 to v2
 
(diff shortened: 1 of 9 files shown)


caleb531:
@YtvwlD Thanks for the PR! But I have one concern, on the note of backwards-compatibility, since I mentioned that was a requirement of mine in https://github.com/caleb531/automata/issues/8#issuecomment-455922942. That is, you currently have `acceptance_mode` as a required parameter. It should be an optional parameter, with the default value set to `'both'` in your function signatures. Otherwise, you would effectively be breaking everyone's existing `DPDA` and `NPDA` instances because they don't have `acceptance_mode` specified.

YtvwlD:
Oh, yeah. Sorry about that. I'll fix it tomorrow.

YtvwlD:
I made `both` the default and added the two tests.
I'm still unsure whether I should add a test which defines the "wrong" mode of acceptance and asserts that the input is not being accepted. This would probably make sense but I would have to remember stuff about PDAs...

caleb531:
@YtvwlD That may be a good idea, and actually wouldn't be difficult. Simply take your existing `final_state` and `empty_stack` tests, swap the acceptance modes, then assert that the input has not been accepted in both cases.

Also, I noticed that the `_validate_acceptance` method is never called, so please add a test for that as well.

screen shot 2019-02-01 at 11 14 13 am

Those will be the final requests I make before I consider this PR perfect. Thank you for your excellent work so far. :)

caleb531:
@YtvwlD As you can already tell, I have taken the liberty of adding these final requests of mine, so you don't have to worry about them. As it is, all tests have passed, and I am ready to merge this PR. If you have any additional work or feedback, throw it into a new issue/PR.

PR #18, opened 2020-12-13
Optimisation for DFA minimisation

eohomegrownapps:
In the breadth first search implementation in `_compute_reachable_states`, when processing the neighbours of a given vertex, for each vertex we check whether it is in `states_checked` and if not push it to the `states_to_check` queue, however, we only add each vertex to `states_checked` once we actually reach that vertex. This could lead to lots of duplication of vertices in the queue, with potentially a O(V * deg(V)) -> worst-case O(V^2) runtime.

Fixing this bug (in other words, adding to the visited set at the same time as we push the vertex to the queue) takes the runtime for minimising ~5000 fairly large DFAs (with around 6 states each after minimisation) from hours to a few minutes.

In addition, as it does not appear to be the case that we need thread-safe processing for the queues here, I changed the queue used from `queue.Queue` to `collections.deque` -- this also yields a performance improvement (see https://stackoverflow.com/a/20330499).

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -3,7 +3,7 @@
 
 import copy
 import itertools
-import queue
+from collections import deque
 
 import automata.base.exceptions as exceptions
 import automata.fa.fa as fa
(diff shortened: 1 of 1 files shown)


caleb531:
@eohomegrownapps This looks fantastic! I never knew that `deque` was actually more performant than `Queue`—a bit ironic, if you ask me. And I would agree: thread-safety is not a requirement for this library.

Because all tests are still passing and code coverage was not affected by your changes, I will absolutely merge in this optimization.

eohomegrownapps:
Thanks for merging this @caleb531 :) On a related note, is the generation of DFA state names based on NFA state names / minimised DFA state names based on unminimised DFA state names implemented this way for pedagogical reasons, and if so would it be worth adding an option to disable this? (again, for very large automata this can lead to very large strings being copied many times...)

caleb531:
@eohomegrownapps The pedagogical argument is certainly one part of it, but even more so, the API implies that each state must have a string name, and I wasn't sure what strategy to use if all the names are auto-generated.

What would you propose an alternative (more scalable) naming system to be? Counting-based? (e.g. 1, 2, 3, 4, etc.)

eohomegrownapps:
@caleb531 I was thinking a counting-based system (e.g. having a dictionary mapping stringified set names to integers and incrementing a count)

PR #23, opened 2021-02-20
Faster minify function

Tagl:
An O(sn log n) implementation of DFA minimization (Hopcroft's algorithm).
Code was originally written by @jaypantone and I copied it with his permission and adapted it to the library.


caleb531:
@Tagl This looks excellent! Thank you for the contribution. Merging now...

PR #37, opened 2021-08-05
Now marking final states in the DOT output for DFAs

progirep:
Currently the graphviz/DOT output does not include marking the final states. This commit changes that.


caleb531:
@progirep This looks great! Thank you for the contribution. Definitely to know which states are the final states.

Issue #40, opened 2021-09-10
Type Annotations


eliotwrobson:
Hello,

Thanks for your work on this library! I've been using it as part of the backend for a course I'm TAing. For our purposes, we're working on adding type hints (which can be checked with mypy) to the library. We think this makes the code more readable, and easier to tell what certain functions do. Would there be interest in including these changes in this project? I'd be happy to put up a pull request once we're done.

Best,
Eliot

caleb531:
@eliotwrobson Maybe... would you be able to guarantee Python 3.6 compatibility? I am reading that Type Annotations have had somewhat of a messy history in Python 3:

https://towardsdatascience.com/type-annotations-in-python-d90990b172dc

What are your thoughts?

eliotwrobson:
We're working in 3.6, so compatibility shouldn't be an issue. We're not really using any super-complex features, just makes it easier to keep track of what's a set, dict, etc.

caleb531:
@eliotwrobson Yeah, I'd say go for it! Please submit me a PR to `master` when you're all done. I'm guessing you won't bother to annotate the tests as well, just the core API modules, correct?

eliotwrobson:
Yes, more than likely just the core API.

caleb531:
@eliotwrobson Great! Then by all means, feel free to start development on this as your own convenience.

eliotwrobson:
I've started some development on this, and I've run into a small issue. There are some internal functions that violate the written interface for the NFA file (in some places, it is assumed states must be strings, but in others, it is assumed they are integers). Should I make changes to the files to respect the interface (and assume states are always strings)? Or is the interface incorrect?

jasonxia17:
@eliotwrobson and I would prefer if the interface supported any hashable type as states. In the algorithms course that we TA for, it often makes sense to use integers or tuples as the states.

caleb531:
@jasonxia17 (cc @eliotwrobson) I agree. Strings are an arbitrary choice: any hashable type should be usable for state names (or symbol names, for that matter). Furthermore, such would represent a backwards-compatible change, since strings are still supported, and the README mostly uses string-state-name examples anyway.

Eliot, I leave it to you if you'd like to declare dedicated interfaces to represent state/symbol names that can be used anywhere in the library (e.g. `StateName` and `SymbolName` interfaces, perhaps).

eliotwrobson:
[...]

PR #43, opened 2021-12-21
[WIP] Adding type hints and code cleanup

eliotwrobson:
Made a first pass at adding type hints to the package (#40). Had to make some more significant (slightly breaking) changes to accommodate some desired features (such as having automaton states be any hashable types). Documentation has not been updated yet, awaiting more discussion from @caleb531 as to whether some of these changes should go in. (cc @jasonxia17)

Also, should be squash + merged and not just merged.


caleb531:
@eliotwrobson Just wanted to follow up on this PR, since it is still in **Draft** status. Are you still working on this? Or is it actually ready for my review?

eliotwrobson:
> @eliotwrobson Just wanted to follow up on this PR, since it is still in **Draft** status. Are you still working on this? Or is it actually ready for my review?

I left this PR in draft status because there were a couple of minor breaking changes that I wasn't sure if you'd like to include. In particular, I wasn't sure how to avoid having to change the name of some of the internal validation functions (mypy doesn't like it if you change the signature of a function in a derived class). I wasn't going to add a test to increase the coverage until you took a look (since I think some of that is related to one of the above breaking changes, and there's only one new uncovered line anyway).

However, in terms of functionality, the code typechecks and all the updated tests pass. Sorry if this isn't how this type of question is supposed to be handled, I wrapped up the initial work and got busy with other things.

caleb531:
@eliotwrobson Hmm, so are you saying that in order to complete this PR, you need to make breaking changes like changing the name of the subclass validation methods?

eliotwrobson:
@caleb531 More or less, I think there _might_ be a way to hack around it (basically tell the typechecker that the overrides can take in any type), but I didn't look into it that closely.

caleb531:
@eliotwrobson Gotcha. Whatever you can do to keep the current API / method names would be preferable, since I chose all the method names very intentionally. But let me know if your alternative approach isn't feasible, and we can reevaluate if that becomes the case.

eliotwrobson:
> @eliotwrobson Just a heads-up: I recently merged #46 which made a bunch of changes across the codebase. I suspect you will need to redo this PR quite a bit if you still have time to work on it.

Hi Caleb, thanks for the heads up. I dropped the ball on this because I wound up spending more time working on improvements to our local version of automata. The more I think about it, I'm not sure it really makes sense to add type hints unless this package is really being used as part of a production system, and the breaking changes I made were really meant to take advantage of having the type hints set up. I'm happy to keep working on this at some point if you think it furthers the goals of the project, but otherwise, I'm not really sure whether it really makes sense to include them. I'm going to go ahead and close this PR, but I can reopen / remake later if you think this is in-line with how you see this package being used.

On a better note, some of the improvements we made might be of interest (for example, we use a significantly faster algorithm for checking DFA equality), and I'd be happy to share those changes with you or commit them as part of other PRs.

caleb531:
@eliotwrobson Understood, and I think I'm inclined to agree re: type hints. I feel like if you have the right tooling in place, many of the types can be inferred anyway.

However, I am very interested in any other improvements you might have to offer (esp. wrt performance), and I would be happy to review any other pull requests you send my way.

eliotwrobson:
> However, I am very interested in any other improvements you might have to offer (esp. wrt performance), and I would be happy to review any other pull requests you send my way. If you do, please set your PR target branch to `develop`, since that is now effectively serving as my "pre-release" branch.

Sounds good. Some of our improvements are rewrites of certain functions, and for those I'll probably just throw up a PR. Others would require introducing a non-trivial dependency, and for those I'll probably create an issue to discuss before making any changes.

PR #48, opened 2022-08-17
Added networkx dependency and faster equality algorithm

### LINKED ISSUE #47: Performance improvement for equality checking
The `__eq__` function in the DFA class uses the symmetric difference between both DFAs, which is an O(n^2) time algorithm (and requires creating an auxiliary DFA). The Hopcroft-Karp algorithm doesn't require creating the auxiliary data structures, but requires the use of a union find. I think it's possible to copy in the code from a generic Python Union-Find, but the `networkx` package has an implementation. Is it too much to introduce this dependency? Any generic union find should work.

eliotwrobson:
Resolves #47. Uses a faster algorithm for checking equality for DFAs, and adds networkx dependency. @caleb531 should be good for review. Kinda unfortunate that you can't really see the performance improvement in the unit tests, but for large DFAs it's definitely noticeable.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -39,7 +39,7 @@ class DFA(fa.FA):
         (state_map_a, state_map_b) = DFA._get_state_maps(self.states, other.states)
 
         # Load new transition dicts
-        new_transitions: DFATransitionsT = dict()
+        new_transitions = dict()
 
         # Transitions of self
         DFA._load_new_transition_dict(state_map_a, self.transitions, new_transitions)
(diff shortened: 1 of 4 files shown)


caleb531:
@eliotwrobson I see you're actively working on this PR. With that, I'm converting it to a Draft until you're ready for me to review and merge it.

BTW, don't worry about all the typo commits. I was planning to merge this as a squash merge to simplify things. Although you are welcome to `git rebase -i` on your own branch if you are comfortable with that.

eliotwrobson:
@caleb531 Just flipped back to open. The changes were just a few things I kept noticing got left in by mistake.

PR #53, opened 2022-08-24
Faster DFA algorithms

eliotwrobson:
Sped up a couple DFA algorithms, and changed some to use comprehension style. Possibly a bit extra, but copied from our implementations of these functions, so figured why not. Also remembered to run flake8 this time.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -464,7 +464,7 @@ class DFA(fa.FA):
     def complement(self):
         """Return the complement of this DFA."""
         new_dfa = self.copy()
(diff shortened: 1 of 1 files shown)


caleb531:
@eliotwrobson This looks great! Thank you for this, and please let me know if you have any other PRs you plan to submit before the upcoming Automata v6 release (no set release date yet, just playing it by ear according to the inflow of contributions from awesome people like yourself).

eliotwrobson:
@caleb531 will do! I think this was the last of our local implementations that differed from what was in the main package, but I'll do another pass at some point in the next week or so.

PR #54, opened 2022-08-25
Even faster comparison algorithms

eliotwrobson:
Added specialized algorithms for subset and disjointness checking. Uses networkx to avoid explicitly constructiong + validating a new DFA to run these comparison functions. In our local repo, this sped up our CI job considerably.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -467,9 +467,32 @@ class DFA(fa.FA):
         new_dfa.final_states ^= self.states
         return new_dfa
 
+    def _get_reachable_states_product_graph(self, other):
+        """Get reachable states corresponding to product graph between self and other"""
+        assert self.input_symbols == other.input_symbols
+
+        # Generate product graph corresponding to component DFAs
+        product_graph = nx.DiGraph([
+            ((start_state_a, start_state_b), (transitions_a[symbol], transitions_b[symbol]))
+            for (start_state_a, transitions_a), (start_state_b, transitions_b) in
+            product(self.transitions.items(), other.transitions.items())
+            for symbol in self.input_symbols
+        ])
+
+        product_initial_state = (self.initial_state, other.initial_state)
+        reachable_states = nx.descendants(product_graph, product_initial_state)
+        reachable_states.add(product_initial_state)
+
+        return reachable_states
+
     def issubset(self, other):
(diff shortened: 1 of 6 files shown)


automata/fa/dfa.py
```diff
@@ -467,17 +467,44 @@ def complement(self):
         new_dfa.final_states ^= self.states
         return new_dfa
 
+    def _get_reachable_states_product_graph(self, other):
+        """Get reachable states corresponding to product graph between self and other"""
+        assert self.input_symbols == other.input_symbols
```
caleb531, line 472: @eliotwrobson I'd prefer to raise some new or existing subclass of `AutomatonException` to handle this check, rather than an `assert`. Can you please add a new `SymbolMismatchError` exception class in `automata/base/exceptions.py` and use that instead? The docstring can be something like `"""The input symbols between the given automata do not match"""`.

To clarify, I'm thinking code like the following:

```python
class SymbolMismatchError(AutomatonException):
    """The input symbols between the given automata do not match"""
    pass
```

```python
if self.input_symbols != other.input_symbols:
  raise SymbolMismatchError('The input symbols between the two given DFAs do not match')
```

Please also update the README to document this new `SymbolMismatchError` exception class.

Thanks,
Caleb

caleb531 (changes requested):
@eliotwrobson Just one requested change, but otherwise looks pretty good!

automata/fa/dfa.py
eliotwrobson: Sounds good! There are a couple of places actually that a similar assert gets used, so I'll go ahead and replace them all.

caleb531: @eliotwrobson Great! Just comment here when you're done with your remaining changes to this PR. It is looking great so far.

eliotwrobson: @caleb531 Just took another look and should be good to go. I realize I keep making incremental changes to the same functions in different PRs. Prepping these changes to migrate over to this repo gave me a chance to see how everything could be optimized even more, but I think eliminating these data structures was the last real hurdle. A nice side effect has been making our CI job significantly more efficient.

caleb531: @eliotwrobson I like the sound of that! Thank for for these contributions. I will merge now.

PR #62, opened 2022-10-07
Regex Parser Refactor

eliotwrobson:
Rewrites the parser for converting regexes to NFAs. Still contains type hints from the base project and comments / docstrings are not updated, but passes the test cases where actual parsing happens. @caleb531figured it was better to have something up before taking the time to polish up all of that to get some comments on the design as early as possible.


automata/parse/postfix.py
```diff
@@ -0,0 +1,132 @@
...
+    while stk:
+        res.append(stk.pop())
+
+    return res
+
+
+def parse_postfix_tokens(postfix_tokens: List[Token[ExpressionResultT]]) -> ExpressionResultT:
+    """Parse list of postfix tokens to produce value of expression"""
+
+    stack: Deque[ExpressionResultT] = deque()
```
caleb531, line 117: @eliotwrobson As much as I personally love deques, is a deque actually faster if the only operations you're performing are `append` (right) and `pop`? I don't see any `appendleft` or `popleft` occurrences in this PR, so was just curious. 😅

automata/fa/nfa.py
```diff
@@ -7,6 +7,7 @@
 from pydot import Dot, Edge, Node
 
 import automata.base.exceptions as exceptions
+from automata.parse.parser import parse_regex
```
caleb531, line 10: @eliotwrobson Is the `automata.parse.parser` module specific to parsing only regexes? If so, I think a less-ambiguous name like `automata.regex.parse` would be better.

If we do this, we should also move `automata.base.regex` to `automata.regex.regex`, so that everything regex-related is under the same sub-package.

caleb531 (changes requested):
@eliotwrobson Overall, looks excellent—thank for you all of this so far. I left some questions for you to review, so please take a look and let me know what you think!

automata/parse/lexer.py
```diff
@@ -0,0 +1,160 @@
...
+        Yields:
+            (token_class, re.Match): all token class whose regexp matches the
+                text, and the related re.Match object.
+        """
+        res = []
+
+        for token_factory_fn, regexp in self._tokens:
+            match = regexp.match(text, pos=start)
+            if match:
+                res.append((token_factory_fn, match))
```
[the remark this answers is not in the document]
eliotwrobson: This function originally yielded, but I switched to a list. On a second look I'll probably go ahead and switch it back (since I think you're right that yielding is faster / more flexible).

automata/parse/lexer.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Returns:
+            (token_kind, token_text): the token kind and its content.
+        """
+        best_token_match = None
+        best_match = None
+
+        for token_factory_fn, match in self.matching_tokens(text, start):
+            if best_match and best_match.end() >= match.end():
+                continue
```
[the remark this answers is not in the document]
eliotwrobson: This is meant to be a continue, though I understand it feels a bit confusing. I'll probably flip the condition and put a break in here, since that feels more natural.

automata/parse/postfix.py
eliotwrobson: Based on this blog post, it appears that deques are faster even when only appending and poping from one side: https://dev.to/v_it_aly/python-deque-vs-listwh-25i9

I think it has to do with the way that deques allocate memory (or something like that), that's more efficient than the way lists do it.

automata/fa/nfa.py
eliotwrobson: Only the `automata.parse.parser` module is specific to regexes, the `lexer` and `postfix` modules are actually completely generic. But I think you're right, it might be best to put this all in a regex sub-package (even if the other components are generic).

automata/parse/postfix.py
caleb531: @eliotwrobson Very well! You have me convinced that deques are the way to go here. 🙂

automata/fa/nfa.py
caleb531: @eliotwrobson Okay, great! Please proceed with that reorg, when you can.

automata/parse/lexer.py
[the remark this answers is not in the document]
caleb531: @eliotwrobson Sounds good to me. Please proceed.

automata/parse/lexer.py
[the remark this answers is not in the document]
caleb531: @eliotwrobson Sounds good, please proceed at your leisure.

eliotwrobson:
@caleb531 addressed your comments and removed helper functions that were used by the old parser. If everything looks good, I'll start polishing things up + porting over our test suite for the individual parser components.

caleb531:
@eliotwrobson Great, thank you! I think the only two outstanding issues were the [`yield` refactor](https://github.com/caleb531/automata/pull/62#discussion_r990364606) and the [`break` refactor](https://github.com/caleb531/automata/pull/62#discussion_r990367589). But if those are not impediments to your migrating the tests, then please go ahead!

tests/test_lexer.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def test_add_token(self) -> None:
+        class AToken(Token):
+            pass
+
+        class BToken(Token):
+            pass
+
+        registry: TokenRegistry = TokenRegistry()
+        assert 0 == len(registry._tokens)
```
caleb531, line 24: @eliotwrobson Can you please convert these to proper unittest assertions (e.g. `self.assertEqual`)?

tests/test_lexer.py
```diff
@@ -0,0 +1,277 @@
...
+        lexer.register_token(lambda x: AAToken(x), r'aa')
+
+        match = lexer.tokens.get_token('aaa')
+        assert match is not None
+
+        token_factory_fn, re_match = match
+        assert isinstance(token_factory_fn('aa'), AAToken)
+
+
+class VerifyRegisterTokensTestCase():
```
caleb531, line 181: @eliotwrobson I think in order for my test runner (`nose2`) to find the tests, the test class must start with `Test` and be subclassed from `unittest.TestCase`. Can you please update this?

caleb531 (changes requested):
@eliotwrobson This is great, but your test classes/methods need to be updated to properly integrate / be discoverable by the test runner.

Thank you for all that you have been doing so far!

caleb531:
@eliotwrobson This is getting pretty close to being ready to merge. The only remaining task I can see is to get the code coverage back up to 100% (through additional/revised tests), if you could please work on that. Once that's done, I'll personally be ready for a final skim-through / merge unless you still have outstanding tasks on your end.

eliotwrobson:
@caleb531 Awesome! The only things that are left on my end are coverage (as you said), cleaning up the docstrings / comments, and removing the type annotations. I'll focus on the tests first (not sure if you want to handle some of that other stuff once this is merged).

caleb531:
@eliotwrobson I trust your coding ability and style, so it would be great if you could take care of the code cleanup part. Besides, nothing personally jumped out at me in your PR as bad style, so I'd prefer to follow your lead with anything you think is worthy of cleanup in that respect.

eliotwrobson:
@caleb531 I actually used a different library as starter code but didn't bother to change the docstrings and other style features, so I definitely want to get those cleaned up before anything gets merged.

caleb531:
>  I actually used a different library as starter code but didn't bother to change the docstrings and other style features, so I definitely want to get those cleaned up before anything gets merged.

@eliotwrobson I see. Which library did you use as starter code? I'm trying to figure out if I it would be proper to update the automata README or LICENSE.txt with any attribution towards that library.

eliotwrobson:
I used the [tdparser library](https://github.com/rbarrois/tdparser).

caleb531:
@eliotwrobson Got it. Would it reduce some of the maintenance burden of this library (automata) if we include [tdparser](https://pypi.org/project/tdparser/) as an official dependency? If there's a lot of overlap with the parsing boilerplate in this PR, I would almost prefer that.

This is all going to roll into the next major release anyway (where breaking changes are okay). But what do you think?

eliotwrobson:
@caleb531 As nice as it would be to lower the maintenance burden, the implementation here has been modified heavily from the initial library, including the parsing algorithm itself and the inheritance structure. This is actually why I went to the trouble of writing this parser, I couldn't find any existing ones that really met our needs. That starter code was really more of a guide for me writing this than anything else.

caleb531:
@eliotwrobson Okay, perfect! I don't think it will be an issue then, if the implementation has diverged so much.

I've got other plans tonight, but I'll check back on this tomorrow in case you've made any more progress. Carry on!

eliotwrobson:
@caleb531 I hope you enjoy your evening! I've uplifted and increased the number of tests, removed the old type hints, and added docstrings and some comments. Everything's passing and should be good to merge!

caleb531:
@eliotwrobson Thanks, it was nice! And great! Although I still see some commits from you this morning. Are you _sure_ this is ready to merge? 😝

eliotwrobson:
@caleb531 Yeah, should be good to go. The changes I made this morning were stylistic inconsistencies I noticed when I was doing a final pass through to do the benchmarking (didn't affect functionality).

caleb531:
@eliotwrobson Great! Will perform the final review and merge now.

I'm looking forward to squashing all these commits lol (reminds me of https://xkcd.com/1296/ 😜):

Screen Shot 2022-10-08 at 10 33 58 AM

caleb531:
@eliotwrobson Just merged your PR! Thanks again for all of this! I did make a few commits directly to your feature branch to fix a coverage issue (not sure why Coveralls was saying 100%, because locally it wasn't). I also simplified some lambda expressions where you could've just passed the function directly.

eliotwrobson:
@caleb531 Awesome, happy to contribute! Thanks for taking the time to review. Working on the open sourced version has given me some insight on changes we can make to our local version before that gets deployed.

caleb531:
@eliotwrobson Oh, that's great to hear! I don't have a specific timeline yet on when I'll be releasing v7 of this library. I typically like to take my time with the major releases so that I can make sure it's polished and has enough newness lol.

caleb531 (commented):
This pull request is already merged, but GitHub thinks I haven't approved these changes yet. Testing to see if I can fix that.

caleb531 (commented):
Testing... no option to approve explicitly, so I think this is as good as it's gonna get. 😅

PR #63, opened 2022-10-09
NFA conversion fix

### LINKED ISSUE #60: Conversion from DFA to NFA does not properly handle state names of mixed data types
I found the following bug:
nfa._add_new_state introduces integers into the set of states.

Current code is:
```
    def _add_new_state(state_set, start=0):
        """Adds new state to the state set and returns it"""
        new_state = start
        while new_state in state_set:
            new_state += 1

        state_set.add(new_state)

        return new_state
```
which could be fixed, for example, in this way:
```
    def _add_new_state(state_set, start=0):
        """Adds new state to the state set and returns it"""
        new_state = start
        while str(new_state) in state_set:
            new_state += 1

        state_set.add(str(new_state))

        return str(new_state)
```
EDIT: this bugfix actually runs into other problems later - this is just an idea
_____________________
This can lead to logic problems later. For example the following code fails:
```
from automata.fa.dfa import DFA
from automata.fa.nfa import NFA

A = NFA(
    states={'0', '1'},
    input_symbols={'0'},
    transitions={'0': {'0': {'1'}}, '1': {'0': {'1'}}},
    initial_state='0',
    final_states={'1'}
)

print(A.reverse().states)

B = DFA.from_nfa(A.reverse())

assert(A.accepts_input('00')==B.accepts_input('00'))
```

eliotwrobson:
Resolves #60

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -4,7 +4,7 @@
 import copy
 from collections import deque
 from enum import IntEnum
-from itertools import product
+from itertools import product, next
 
 import networkx as nx
 from pydot import Dot, Edge, Node
@@ -548,17 +548,11 @@ class DFA(fa.FA):
 
         # Data structures for state renaming
         new_state_name_dict = dict()
-        state_name_counter = 0
+        state_name_counter = count(0)
+
         def get_name_renamed(states):
             nonlocal state_name_counter, new_state_name_dict
-            canonical_form_states = frozenset(states)
-
-            if canonical_form_states in new_state_name_dict:
-                return new_state_name_dict[canonical_form_states]
-
-            new_state_name_dict[canonical_form_states] = state_name_counter
-            state_name_counter += 1
-            return state_name_counter-1
+            return new_state_name_dict.setdefault(frozenset(states), next(state_name_counter))
 
         def get_name_original(states):
             return frozenset(states)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -4,7 +4,7 @@
 import copy
 from collections import deque
 from enum import IntEnum
-from itertools import product, next
+from itertools import product, count
 
 import networkx as nx
 from pydot import Dot, Edge, Node
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -545,7 +545,7 @@ class DFA(fa.FA):
         return tuple(sorted(states, key=str))
 
     @classmethod
-    def from_nfa(cls, target_nfa, retain_names):
+    def from_nfa(cls, target_nfa, retain_names=False):
         """Initialize this DFA as one equivalent to the given NFA."""
         dfa_states = set()
         dfa_symbols = target_nfa.input_symbols
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -539,11 +539,6 @@ class DFA(fa.FA):
         """Stringify the given set of states as a single state name."""
         return '{{{}}}'.format(','.join(sorted(str(state) for state in states)))
 
-    @staticmethod
-    def _to_canonical_form(states):
-        """Return a canonical (hashable) form of the given iterable of states."""
-        return tuple(sorted(states, key=str))
-
     @classmethod
     def from_nfa(cls, target_nfa, retain_names=False):
         """Initialize this DFA as one equivalent to the given NFA."""
@@ -556,7 +551,7 @@ class DFA(fa.FA):
         state_name_counter = 0
         def get_name_renamed(states):
             nonlocal state_name_counter, new_state_name_dict
-            canonical_form_states = cls._to_canonical_form(states)
+            canonical_form_states = frozenset(states)
 
             if canonical_form_states in new_state_name_dict:
                 return new_state_name_dict[canonical_form_states]
@@ -566,7 +561,7 @@ class DFA(fa.FA):
             return state_name_counter-1
 
         def get_name_original(states):
-            return cls._to_canonical_form(states)
+            return frozenset(states)
 
         get_name = get_name_original if retain_names else get_name_renamed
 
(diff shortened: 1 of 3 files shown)


eliotwrobson:
@caleb531 One thing to note is that this makes the DFA algorithms kindof inconsistent, since the others always convert merged state names to strings but `from_nfa` leaves them as tuples.

caleb531:
> One thing to note is that this makes the DFA algorithms kindof inconsistent, since the others always convert merged state names to strings but from_nfa leaves them as frozensets.

@eliotwrobson Sure, but `NFA.from_dfa` was already creating state names (for the new NFA) as integers, regardless of the type of the original DFA's state names. Therefore, I suspect it's fine for the data type of the state names to change if the type of the instance is changing. Conversely, if the instance type is not changing, then the type of the state names should not change.

Going a step further, I wonder if it might be a good convention to enforce (in the `validate` method) that each individual `Automaton` sub-instance should not mix data types for its state names—they must all use the same data type. Kinda like the Python rule stipulating that you cannot mix spaces and tabs—it's an either/or choice (disregarding for a moment that PEP 8 prefers 4 spaces always, but you get my point).

In practice, this means that a DFA/NFA instance can only have state names of a single data type (e.g. str only, or int only, but never str mixed with int).

Would such a rule invalidate the need for this PR? Maybe, since the original situation from which this PR arose was your (somewhat contrived) example:

```python
A = NFA(
    states={1, '1'},
    input_symbols={'0'},
    transitions={1: {'0': {'1'}}, '1': {'0': {'1'}}},
    initial_state=1,
    final_states={'1'}
)

B = DFA.from_nfa(A)

assert A.accepts_input('00') == B.accepts_input('00')
```

But, if we stipulate that the above NFA must have either state names of all `str` type, OR state names of all `int` type, then this issue goes away, does it not?

Then again, tuples would probably be better than rigid string values, so I may still choose to accept this PR. But to your original point about the inconsistency, I'm thinking that the state name type is welcome to change when converting to another type of automaton (i.e. DFA->NFA or vice-versa).

Okay, thanks for listening to my sermon 😂😅. Thoughts?

eliotwrobson:
You definitely bring up some valid points about consistency. In practice, I'm not sure how enforceable that convention is (for example, if state names can be tuples, do we enforce they all have the same length and same component types?).

In our fork, I already made the change to have all the DFA functions use tuples or frozensets for output state names, allowing mixed types for state names. This is because a bunch of the DFAs people construct in our course have mixed types for state names (since this is how a lot of our solutions are written). I'd recommend making the same change here, since it allows for more flexibility than enforcing one type for state names.

Although I raised the issue of inconsistency, I actually have a version of the tests using the tuples / frozensets as described (instead of converting to strings). I'd be happy to open a PR for that if you accept this one and go in that direction.

caleb531:
@eliotwrobson I agree. But if we're going to do that, then we should also rewrite my recently-updated `NFA._add_new_state` (and possibly also `NFA.from_dfa`) to ensure that values like `1` (int) and `'1'` (str) are **not** considered equal.

Unless you think that would be too much of a footgun for people who might forget the quotes, but if we're moving away from string coercion, then I'd argue `1` and `'1'` _must_ be considered separate / not-equal.

eliotwrobson:
> then we should also rewrite my recently-updated `NFA._add_new_state` (and possibly also `NFA.from_dfa`) to ensure that values like `1` (int) and `'1'` (str) are **_not_** considered equal.

I'm in agreement that the string and int versions should not be considered equal. I don't think it's that much of a footgun since it's more about setting a consistent convention than anything else (and the validation functions are there to prevent making inconsistent FAs).

I don't think these rewrites should be a big deal (and am happy to take them on), since I think we can just revert `NFA._add_new_state` to what it was before the rewrite, and `from_dfa` _should_ just work as written. Our fork is written with these assumptions in mind, so it's just a matter of copying things over on an as-needed basis.

caleb531:
@eliotwrobson Okay, great! I was expecting the fix for `NFA._add_new_state` would effectively be reverting my changes to the method (you probably don't want to revert the entire commit dd9036683c9e46cde387bd0b98d5d3f4ebbb89b5 though, since it includes the additional test case for mixed state name types).

Thank you for your continued kindness in contributing so much code to this core library! I need to buy you a pizza or something. But yes, I'll gladly review any more PRs you send my way! 😄

Speaking of which, given our conversation here, did you have any more changes to make to this one before it's ready for my review?

eliotwrobson:
@caleb531 no more changes to this one! I'll go ahead and take it out of draft. Once this is merged I'll put up the PR with the other rewrites.

automata/fa/dfa.py
```diff
@@ -540,44 +540,61 @@ def _stringify_states(states):
...
-        dfa_transitions = dict()
+        dfa_transitions = {}
+
+        # Data structures for state renaming
+        new_state_name_dict = dict()
+        state_name_counter = count(0)
+
+        def get_name_renamed(states):
+            nonlocal state_name_counter, new_state_name_dict
+            return new_state_name_dict.setdefault(frozenset(states), next(state_name_counter))
```
caleb531, line 555: @eliotwrobson This looks a lot like what I was _trying_ to do with `NFA._add_new_state`, but just much better executed lol.

Is there a reason you chose to use `itertools.count` as the counter variable, rather than just an `int` value and `+=` incrementing logic? I've never used `itertools.count` before, so I'm mostly just curious what use cases it was created to solve.

eliotwrobson: Itertools count is just for situations where you want to be able to increment an integer counter and return the value of that counter each time (i.e. exactly this situation). This function originally used an int value with incrementing logic, but using `count` and calling the `next` function is basically just to get a one-line solution (which I realized you could do as I was copying it over).

caleb531: @eliotwrobson I mean, it's certainly cleaner, and seemingly more declarative too (e.g. no fussing about whether the `+=` line happens _before_ or _after_ where the retrieval occurs). I like it!

PR #64, opened 2022-10-09
Other rewrites to remove string conversion

eliotwrobson:
Removes string conversion in other functions. @caleb531 new functions should be good to go.


tests/test_dfa.py
```diff
@@ -977,90 +977,93 @@ def test_minify_dfa_complex(self):
                           '48', '22', '2', '36', '34', '49', '40', '25', '30',
                           '53', '42', '33', '19', '50', '29'})
         check_dfa = DFA(
-            states={'5', '36', '1', '49', '40', '25', '46', '6', '55',
-                    '{0,10,14,17,18,23,26,28,31,37,4,41,45,47,56,8}', '33',
-                    '11', '20', '48', '44', '32'},
+            states={frozenset(('5',)), frozenset(('36',)), frozenset(('1',)), frozenset(('49',)), frozenset(('40',)),
```
caleb531, line 980: @eliotwrobson In practice, will your algorithm changes still work even if developers opt to use non-frozen set literals instead of `frozenset`?

automata/fa/dfa.py
```diff
@@ -318,15 +318,9 @@ def _merge_states(self, retain_names=False):
...
-        def get_name(eq, i):
-            if retain_names:
-                return list(eq)[0] if len(eq) == 1 else DFA._stringify_states(eq)
-
-            return str(i)
-
         # now eq_classes are good to go, make them a list for ordering
         eq_class_name_pairs = [
-            (eq, get_name(eq, i))
+            (eq, frozenset(eq) if retain_names else str(i))
```
caleb531, line 323: @eliotwrobson Not trying to [pre-optimize](https://stackify.com/premature-optimization-evil/) here 😅, but do you think it would be more performant for the `retain_names` check to be _outside_ the loop so it only runs once?

```python
if retain_names:
    eq_class_name_pairs = [
        (eq, frozenset(eq))
        for eq in eq_classes
    ]
else:
    eq_class_name_pairs = [
        (eq, str(i))
        for i, eq in enumerate(eq_classes)
    ]
```

caleb531 (changes requested):
@eliotwrobson Just a few questions, but otherwise looking great!

automata/fa/dfa.py
eliotwrobson: Probably doesn't make a huge difference but I'll go ahead and rewrite this line.

tests/test_dfa.py
eliotwrobson: What do you mean by "non-frozen set literals"? This is just asserting that the algorithm outputs frozensets for this test case.

caleb531: @eliotwrobson By non-frozen set literals, I was referring to the literal set syntax, e.g. `{1, 2, 3, ...}`. But I _just_ realized that wouldn't work anyway, since mutable sets are not hashable (i.e. you can't have a mutable set inside of another set).

So sorry about that—I was just overthinking it. 😅

automata/fa/dfa.py
eliotwrobson: @caleb531 did a mini-rewrite and looks cleaner now!

caleb531 (approved):

PR #66, opened 2022-10-13
Optimized the DFA.minify() algorithm

eliotwrobson:
Modifications that significantly speed up the minify algorithm. Will post benchmarks soon, but seeing 10x speedup in some early tests. @caleb531 Should be mostly complete, will do one more pass before merging. Figured it was better to have a version up ASAP rather than waiting for things to be flawless.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -4,14 +4,14 @@
 import copy
 from collections import deque
 from enum import IntEnum
-from itertools import count, product, chain
+from itertools import chain, count, product
 
 import networkx as nx
 from pydot import Dot, Edge, Node
 
 import automata.base.exceptions as exceptions
-from automata.base.utils import PartitionRefinement
 import automata.fa.fa as fa
+from automata.base.utils import PartitionRefinement
 
 
 class OriginEnum(IntEnum):
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -353,10 +353,10 @@ class DFA(fa.FA):
             for active_letter in self.input_symbols:
 
                 origin_dict = transition_back_map[active_letter]
-                states_that_move_into_active_state_chain = chain(*(
+                states_that_move_into_active_state_chain = chain.from_iterable(
                     origin_dict[end_state]
                     for end_state in eq_classes.get_set_by_id(active_state_id)
-                ))
+                )
 
                 # Using a tuple because we don't need to make a deep copy
                 new_eq_class_pairs = eq_classes.refine(states_that_move_into_active_state_chain)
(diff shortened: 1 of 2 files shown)


eliotwrobson:
Old minimization: 
image

New minimization:
image

automata/fa/dfa.py
```diff
@@ -18,6 +18,62 @@ class OriginEnum(IntEnum):
     OTHER = 1
 
 
+class PartitionRefinement:
```
caleb531, line 21: @eliotwrobson Is this a class that can be leveraged by other primitives in this library (e.g. NFA)? Because if so, I think it would be better to move to a separate file.

caleb531 (changes requested):
@eliotwrobson Looks great—just one question about the `PartitionRefinement` class.

automata/fa/dfa.py
eliotwrobson: It's probably worth moving to its own file anyway (since this class can be used independently of the DFA). What subheading should this be in? Just under `automata.base`?

I've actually never seen this data structure used before outside of this specific algorithm. Still probably should go in its own file, but interesting that it doesn't come up very often.

caleb531: @eliotwrobson If we go that route, I would suggest the module path be `automata.base.partition_refinement`.

BUT, I was also considering if a `automta.base.utils` module would be better. If we go with this approach, we could also move `OriginEnum` there so that also becomes reusable. Thoughts?

eliotwrobson: I think moving the enum there is a bit overkill. It's very short and only really usable in the equality algorithm. I actually considered making the definition local to that function.

caleb531: @eliotwrobson Yeah, I can see that. But regardless, I think I'd still like to move `PartitionRefinement` under a `automata.base.utils` module. Because it's really just a utility class, and I don't want it to be seen as having the same significance the other library primitives like `.fa.dfa` or `.fa.nfa` or even `.base.automaton`. I also want to create a future-proof place to put ancillary data structures like these.

eliotwrobson: Sounds good! I'll go ahead and move it when I'm back at my desk.

eliotwrobson: @caleb531 class has been moved!

caleb531:
@eliotwrobson This is looking perfect now! Were there any other tweaks you wanted to make before I merge?

eliotwrobson:
@caleb531 Don't think so! I'm adding this to our library so if I uncover something really strange I'll put up another PR, but otherwise it's looking good and the benchmarks have been excellent.

caleb531:
@eliotwrobson Okay, great! BTW, for future PRs, please also run `isort` on the files so that the import order is more PEP8-compliant.

I'll commit the tweak directly to your fork as far as this PR is concerned.

caleb531 (approved):

PR #67, opened 2022-10-14
Fixed a bug with the partition algorithm

eliotwrobson:
Sorry! 😅  While integrating the code into our codebase, noticed a subtle bug that caused the tests to fail very infrequently (like 1 out of every 20 runs). Also, cleaned up variable names to more closely match the wikipedia article: https://en.wikipedia.org/wiki/DFA_minimization#Hopcroft's_algorithm

### DIFF
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -10,6 +10,8 @@ class PartitionRefinement:
     Adapted from code by D. Eppstein: https://www.ics.uci.edu/~eppstein/PADS/PartitionRefinement.py
     """
 
+    __slots__ = ["_sets", "_partition"]
+
     def __init__(self, items):
         """Create a new partition refinement data structure for the given
         items. Initially, all items belong to the same subset.
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -294,26 +294,26 @@ class DFA(fa.FA):
         processing = {final_states_id}
 
         while processing:
-            active_state_id = processing.pop()
+            # Save a copy of the set, since it could get modifief while executing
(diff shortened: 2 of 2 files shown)


caleb531:
@eliotwrobson Oh interesting, I haven't noticed this. Do you have any idea 

At first, I was thinking the intermittency of the failures could have something to do with the unordered nature of Python sets, or hash randomization. But looking at the code, it's difficult to say for sure.

What do you think? Can you think of any test case you could add to the suite that would cause the absence of this fix to fail 100% of the time?

eliotwrobson:
@caleb531 I can take another look a little bit later. The error comes up because the active set from the processing set can get modified in the body of that outer loop. But you're right, the reason it comes up intermittently is because of the non-deterministic behavior of the set data structures that the algorithm is iterating over. This is actually why this was difficult to debug, I had to switch the data structures to be deterministic to have it come up regularly.

With respect to test cases, you can actually encounter it even with small DFAs, but the non-determinism makes it very hard to reproduce. I only really found it because our test suite has parameterized versions of some of the tests here (runs similar tests with many different DFAs configured as fixtures).

caleb531:
@eliotwrobson Well in that case, I trust your judgement. It probably wouldn't be very feasible to test due to, as you said, the non-determinism of the data structures involved.

In that case, I will proceed to merge this PR, but as always, feel free to send any additional PRs my way if you find anything else.

eliotwrobson:
@caleb531 Sounds good. Yeah, unfortunately this is one of those where we only really found the issue by just having a huge number of test cases.

PR #68, opened 2022-10-16
Remove concat token from parser and fix old comment

eliotwrobson:
@caleb531 Replaces an out-of-date comment and removes the concat token from the regex parser (since the concatenation operator is implicit and not part of the syntax in the documentation).


caleb531 (approved):
@eliotwrobson LGTM! 😀

PR #69, opened 2022-10-16
Remove unnecessary function in parser

eliotwrobson:
@caleb531 Removes redundant function in constructor of a class in parser.


caleb531 (approved):
Looks great! Thank you for this!

PR #70, opened 2022-10-18
Remove some set conversions in DFA.from_nfa()

eliotwrobson:
Removes some redundant set conversions in DFA.from_nfa function. Noticeable performance improvement in some benchmarks.


caleb531 (approved):
@eliotwrobson Looks great! Thank you for this.

PR #73, opened 2022-10-20
Switch GNFA functions to respect immutability

eliotwrobson:
@caleb531 see title. Functions are a bit confusing so could maybe be refactored more, but a solid start I think.


caleb531 (approved):
Eh, I wouldn't worry about the complexity too much. It's fine for now. 🙂

PR #75, opened 2022-10-22
Optimize reachable states algorithms

eliotwrobson:
@caleb531 turns out it's faster in some cases to do BFS directly instead of reading all states of a product graph into a data structure (since this effectively makes two passes through the data instead of 1). Networkx implements the same algorithm anyway in pure Python. Saw 10x speedup in some cases.


caleb531 (approved):
Looks good to me. Obviously, not as short and sweet as the networkx algo, but performance is more important than that.

PR #78, opened 2022-10-23
First attempt at writing BFS-based NFA subset

eliotwrobson:
Currently failing tests, but first attempt at implementation. Tried to implement the construction from https://www3.nd.edu/~dchiang/teaching/theory/2020/hw/hw2.pdf and use this for determining subsets.

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Classes and methods for working with nondeterministic finite automata."""
 from collections import deque
-from itertools import chain
+from itertools import chain, product
 
 import networkx as nx
 from frozendict import frozendict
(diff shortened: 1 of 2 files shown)


eliotwrobson:
@caleb531 So turns out there's no efficient algorithm for this, aside from effectively turning both NFA's into DFA's (I'm not exactly sure how but I think my attempt at doing this was flawed): https://cs.stackexchange.com/questions/9954/is-there-an-efficient-test-for-if-an-nfa-accepts-a-subset-of-another-nfa

While it _might_ be possible to do this, I think the algorithm will ultimately boil down to converting both NFAs to DFAs and running that algorithm.

caleb531:
@eliotwrobson That's fine! Could you possibly submit a PR with that DFA-based subset check for NFA?

I'm currently staring into the abyss of refactoring MNTM to be immutable (as it has some additional internal state and complexity which must be unwound to be properly immutable). Pray for me. 😂😅

eliotwrobson:
@caleb531 I could, but I'm so sure it's worth doing. The NFA -> DFA conversion is quite expensive, and I don't think it's good practice to have very expensive algorithms run under the hood without giving the end user fine-grained control over what's happening. For example, having a method that does the subset check means the end user can't do anything with the resulting equivalent DFAs after (and thus, makes it easy to accidentally perform this expensive computation twice). 

Good luck with the MNTM! Reconciling the immutability with the automata that have large internal state definitely seems like one of the trickier aspects of this change.

caleb531:
> For example, having a method that does the subset check means the end user can't do anything with the resulting equivalent DFAs after (and thus, makes it easy to accidentally perform this expensive computation twice).

@eliotwrobson You have a very good point here. And I trust your judgment if you think there's a sufficiently-efficient algorithm still out there.

And thanks! It's coming along slowly but surely.

Issue #79, opened 2022-10-23
Implement Automaton __iter__ for enumerating accepted language?


caleb531:
@eliotwrobson So with all the recent discussion around automata ultimately being representations of finite languages, it got me thinking:

What if we implemented an `__iter__` method that produced each string in that automaton's accepted language? My thought it we could generate one string at a time via `yield`, and leave it up to the user if they want to convert the automaton to a `list` or `set`. Ideally, we could start with DFAs/NFAs, and then explore if a similar algorithm is possible with PDAs and TMs.

How difficult do you think this would be to implement from an algorithmic perspective? Initially, my thought is generating all permutations of the machine's input symbols, then filter that set down to only those accepted via `accepts_input`.

Although the number of iterations could be enormous, so maybe we could filter that set further by analyzing transition symbols from the initial state (first character) and final states (possible last characters). But I'm curious if there is an approach that isn't so brute-force.

eliotwrobson:
@caleb531 for DFA/NFAs, I think there are algorithms for this based on treating the transitions as a graph and doing graph search (or similar algorithms). I'll have to look through some books to see if they explicitly describe something like this, but here's a stackoverflow post describing something similar:
 https://stackoverflow.com/questions/33058142/optimal-algorithm-to-count-the-number-of-strings-a-dfa-accepts

As for the other automation, I'm not sure about PDAs, but for TMs even checking whether a TM accepts _any_ string (let alone enumeration of all of them) is undecidable (equivalent to the halting problem).

Tagl:
I think I have some code for this already written. I'll take a look tomorrow.

Tagl:
Decided to look at it before I go to bed.
So this is essentially just a dynamic programming problem.
We can define recursive functions and then we just have to store results in memory to cut our time complexity down to $\mathcal{O}(nk)$, and raising space complexity to the $\mathcal{O}(nk)$, where $n$ is the number of states and $k$ is the length of words.
Of course for generating the words instead of just counting them we have to account for constructing each string as well.
This will add *at least* a factor of $k$ to the time complexity and a factor of $k$ to space complexity, but state space remains $nk$.
I'll finish the code and optimize after work tomorrow but here's essentially the logic:
```python
    def count_words_of_length(self, k):
        """
        Counts the number of words of length k in the language represented by the DFA
        """
        return self._count_words_of_length(self.initial_state, k)

    def _count_words_of_length(self, state, k):
        """
        Counts words of length k assuming state is the initial state
        """
[...]

PR #80, opened 2022-10-25
Enumeration and iteration for DFA

Tagl:
Code for #79 
I had to update the copy function in automaton to skip copying the caches. Don't really like the fix I made.
Rest is just as discussed and then added min and max word length functions which are calculated with BFS and longest path in DAG, respectively.

### DIFF
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -115,6 +115,6 @@ class Automaton(metaclass=abc.ABCMeta):
             for attr_name, attr_value in self.attributes().items())
         return f'{self.__class__.__qualname__}({values})'
 
-    def __contains__(self, other):
+    def __contains__(self, input_str):
         """Returns whether the word is accepted by the automaton."""
-        return self.accepts_input(other)
+        return self.accepts_input(input_str)
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -112,5 +112,5 @@ class Automaton(metaclass=abc.ABCMeta):
         """Return a string representation of the automaton."""
         values = ', '.join(
             f'{attr_name}={self._get_repr_friendly_value(attr_value)!r}'
-            for attr_name, attr_value in self.__dict__.items())
+            for attr_name, attr_value in self.attributes().items())
         return f'{self.__class__.__qualname__}({values})'
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -79,11 +79,15 @@ class Automaton(metaclass=abc.ABCMeta):
                 'final states are not valid ({})'.format(
                     ', '.join(str(state) for state in invalid_states)))
 
+    def attributes(self):
+        """Return the public attributes for this automaton."""
+        return {key: value
+                for key, value in vars(self).items()
+                if not key.startswith('_')}
+
     def copy(self):
         """Create a deep copy of the automaton."""
-        return self.__class__(**{key: value
-                                 for key, value in vars(self).items()
-                                 if not key.startswith('_')})
+        return self.__class__(**self.attributes())
 
     # Format the given value for string output via repr() or str(); this exists for the purpose of displaying
 
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -114,3 +114,7 @@ class Automaton(metaclass=abc.ABCMeta):
             f'{attr_name}={self._get_repr_friendly_value(attr_value)!r}'
             for attr_name, attr_value in self.attributes().items())
         return f'{self.__class__.__qualname__}({values})'
+
+    def __contains__(self, other):
+        """Returns whether the word is accepted by the automaton."""
+        return self.accepts_input(other)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -648,7 +648,7 @@ class DFA(fa.FA):
             self._word_cache.append(defaultdict(list))
             level = self._word_cache[i]
             if i == 0:
-                level.update({state: {''} for state in self.final_states})
+                level.update({state: [''] for state in self.final_states})
             else:
                 prev_level = self._word_cache[i-1]
                 level.update({
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -680,7 +680,7 @@ class DFA(fa.FA):
         distances = defaultdict(lambda: float('inf'))
         distances[self.initial_state] = 0
         queue.append(self.initial_state)
-        while len(queue):
+        while queue:
             state = queue.popleft()
             if state in self.final_states:
                 return distances[state]
@@ -712,11 +712,6 @@ class DFA(fa.FA):
             return nx.dag_longest_path_length(subgraph)
         except nx.exception.NetworkXUnfeasible:
             return float('inf')
-
-    @staticmethod
-    def _stringify_states_unsorted(states):
-        """Stringify the given set of states as a single state name."""
-        return '{{{}}}'.format(','.join(states))
 
     @classmethod
     def from_finite_language(cls, language, input_symbols):
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -669,13 +669,27 @@ class DFA(fa.FA):
         """
         Returns the length of the shortest word in the language represented by the DFA
         """
-        return 0
+        if self.isempty():
+            raise ValueError('The language represented by the DFA is empty')
+        queue = deque()
+        distances = defaultdict(lambda: float('inf'))
+        distances[self.initial_state] = 0
+        queue.append(self.initial_state)
+        while len(queue):
+            state = queue.popleft()
+            if state in self.final_states:
+                return distances[state]
+            for _, next_state in self.transitions[state].items():
+                if distances[next_state] == float('inf'):
+                    distances[next_state] = distances[state] + 1
+                    queue.append(next_state)
+        assert False
 
     def maximum_word_length(self):
         """
         Returns the length of the longest word in the language represented by the DFA
         """
-        if self.isfinite():
+        if not self.isfinite():
             return float('inf')
         return 100
 
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -600,21 +600,9 @@ class DFA(fa.FA):
         """
         Returns True if the DFA accepts a finite language, False otherwise.
         """
-        G = self._get_digraph()
-
-        accessible_nodes = nx.descendants(G, self.initial_state) | {self.initial_state}
-
-        coaccessible_nodes = self.final_states.union(*(
-            nx.ancestors(G, state)
-            for state in self.final_states
-        ))
-
-        important_nodes = accessible_nodes.intersection(coaccessible_nodes)
-
         try:
-            nx.find_cycle(G.subgraph(important_nodes))
-            return False
-        except nx.exception.NetworkXNoCycle:
+            return self.maximum_word_length() != float('inf')
+        except ValueError:
             return True
 
     def count_words_of_length(self, k):
@@ -674,8 +662,6 @@ class DFA(fa.FA):
         """
         Returns the length of the shortest word in the language represented by the DFA
         """
-        if self.isempty():
-            raise ValueError('The language represented by the DFA is empty')
         queue = deque()
         distances = defaultdict(lambda: float('inf'))
         distances[self.initial_state] = 0
@@ -688,7 +674,7 @@ class DFA(fa.FA):
                 if distances[next_state] == float('inf'):
                     distances[next_state] = distances[state] + 1
                     queue.append(next_state)
-        assert False
+        raise ValueError('The language represented by the DFA is empty')
 
     def maximum_word_length(self):
         """
@@ -696,8 +682,6 @@ class DFA(fa.FA):
         """
         if self.isempty():
             raise ValueError('The language represented by the DFA is empty')
-        if not self.isfinite():
-            return float('inf')
         G = self._get_digraph()
 
         accessible_nodes = nx.descendants(G, self.initial_state) | {self.initial_state}
@@ -708,8 +692,11 @@ class DFA(fa.FA):
         ))
 
         important_nodes = accessible_nodes.intersection(coaccessible_nodes)
-
-        return nx.dag_longest_path_length(G.subgraph(important_nodes))
+        subgraph = G.subgraph(important_nodes)
+        try:
+            return nx.dag_longest_path_length(subgraph)
+        except nx.exception.NetworkXUnfeasible:
+            return float('inf')
 
     @staticmethod
     def _stringify_states_unsorted(states):
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -140,10 +140,6 @@ class DFA(fa.FA):
         """Return the complement of this DFA and another DFA."""
         return self.complement()
 
-    def __contains__(self, other):
-        """Returns whether the word is accepted by the DFA."""
-        return self.accepts_input(other)
-
     def __iter__(self):
         """
         Iterates through all words in the language represented by the DFA.
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -149,10 +149,11 @@ class DFA(fa.FA):
         Iterates through all words in the language represented by the DFA.
         The words are ordered first by length and then by the order of the input symbol set.
         """
-        min_len = self.minimum_word_length()
-        max_len = self.maximum_word_length()
-        for i in range(min_len, max_len + 1):
+        i = self.minimum_word_length()
+        limit = self.maximum_word_length()
+        while i <= limit:
             yield from self.words_of_length(i)
+            i += 1
 
     def __len__(self):
         """Returns the cardinality of the language represented by the DFA."""
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -689,9 +689,22 @@ class DFA(fa.FA):
         """
         Returns the length of the longest word in the language represented by the DFA
         """
+        if self.isempty():
+            raise ValueError('The language represented by the DFA is empty')
         if not self.isfinite():
             return float('inf')
-        return 100
+        G = self._get_digraph()
+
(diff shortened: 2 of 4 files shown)


caleb531:
@Tagl Thanks so much for this! Will take a closer look at the code tomorrow.

automata/fa/dfa.py
```diff
@@ -529,6 +548,93 @@ def isfinite(self):
...
+            i = len(self._word_cache)
+            self._word_cache.append(defaultdict(set))
+            if i == 0:
+                for state in self.final_states:
+                    self._word_cache[i][state].add('')
+            else:
+                for state in self.states:
+                    for symbol, suffix_state in self.transitions[state].items():
+                        for word in self._word_cache[i-1][suffix_state]:
+                            self._word_cache[i][state].add(symbol + word)
```
eliotwrobson, line 595: This entry in the cache is accessed a lot, and `i` is not modified in this loop. You should factor out some of these lookups, as they can get expensive.

eliotwrobson (commented):
Looks good! A few notes on efficiency. Also, @caleb531 since some of the information is shared between functions here and existing ones (like generating the digraph and stuff), wonder if those should be made cached properties. Even for Python 3.7, we can use a back-port: https://pypi.org/project/cached-property/

caleb531:
@eliotwrobson You know, it's funny you mention `@cached_property`, because before the whole immutability discussion, I was going to implement that for `NFA.lambda_closures`. But that was before I realized cached properties were only supported natively in Python 3.9 or newer. Fortunately, we don't need that anymore in the case of `lambda_closures` because NFAs are now immutable anyway.

But thinking about it more, I'm fine with adding the backport to this library so we can leverage cached properties in places where they're really needed. Especially since `@lru_cache` (which _is_ available in Python 3.7) is [not friendly to class instances](https://rednafi.github.io/reflections/dont-wrap-instance-methods-with-functoolslru_cache-decorator-in-python.html).

automata/fa/dfa.py
Tagl: Good point, updated

automata/fa/dfa.py
```diff
@@ -529,6 +548,93 @@ def isfinite(self):
...
+
+    def minimum_word_length(self):
+        """
+        Returns the length of the shortest word in the language represented by the DFA
+        """
+        if self.isempty():
```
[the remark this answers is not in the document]
Tagl: Agreed. With immutability we will want to cache the results of some of these functions such as isempty, isfinite, maximum_word_length and minimum_word_length.

automata/fa/dfa.py
```diff
@@ -529,6 +548,93 @@ def isfinite(self):
...
+                    queue.append(next_state)
+        assert False
+
+    def maximum_word_length(self):
+        """
+        Returns the length of the longest word in the language represented by the DFA
+        """
+        if self.isempty():
+            raise ValueError('The language represented by the DFA is empty')
+        if not self.isfinite():
```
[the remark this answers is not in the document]
Tagl: I now only construct it in `maximum_word_length` and then `isfinite` calls `maximum_word_length`.

tests/test_dfa.py
```diff
@@ -1308,3 +1308,162 @@ def test_show_diagram_write_file(self):
         self.dfa.show_diagram(path=diagram_path)
         self.assertTrue(os.path.exists(diagram_path))
         os.remove(diagram_path)
+
```
eliotwrobson, line 1311: Maybe not totally needed, but I merged a function that creates a minimal DFA accepting a finite language, so a cool test would be to feed in a finite language to that and check that your code here is able to recover it.

automata/fa/dfa.py
```diff
@@ -512,6 +531,88 @@ def isfinite(self):
...
+    def _ensure_words_of_length(self, k):
+        """
+        Populate word cache up to length k
+        """
+        while len(self._word_cache) <= k:
+            i = len(self._word_cache)
+            self._word_cache.append(defaultdict(set))
+            level = self._word_cache[i]
+            if i == 0:
+                for state in self.final_states:
```
eliotwrobson, line 582: Maybe a bit extra, but I think you can do the inside of both of the loops with comprehensions.

tests/test_dfa.py
Tagl: Agreed

automata/fa/dfa.py
```diff
@@ -512,6 +531,88 @@ def isfinite(self):
...
+            self._count_cache.append(defaultdict(int))
+            level = self._count_cache[i]
+            if i == 0:
+                for state in self.final_states:
+                    level[state] = 1
+            else:
+                prev_level = self._count_cache[i-1]
+                for state in self.states:
+                    state_count = 0
+                    for _, suffix_state in self.transitions[state].items():
```
eliotwrobson, line 561:
```suggestion
                    for suffix_state in self.transitions[state].values():
```

Tagl: Fixed

automata/fa/dfa.py
Tagl: Like this? Or you had something else in mind?

automata/base/automaton.py
```diff
@@ -67,7 +67,7 @@ def _validate_final_states(self):
 
     def copy(self):
         """Create a deep copy of the automaton."""
-        return self.__class__(**vars(self))
+        return self.__class__(**{k: v for k, v in vars(self).items() if not k.startswith('_')})
```
caleb531, line 70: @Tagl Great idea to exclude the private-marked attributes.

automata/fa/dfa.py
```diff
@@ -148,6 +150,23 @@ def __invert__(self):
         """Return the complement of this DFA and another DFA."""
         return self.complement()
 
+    def __contains__(self, other):
```
caleb531, line 153: @Tagl Could you please add docstrings to this and the other magic methods below?

automata/fa/dfa.py
```diff
@@ -148,6 +150,23 @@ def __invert__(self):
...
+        return self.accepts_input(other)
+
+    def __iter__(self):
+        i = self.minimum_word_length()
+        limit = self.maximum_word_length()
+        while i <= limit:
```
caleb531, line 159: @Tagl I would much prefer a `for` loop here to eliminate the manual counting:

```python
def __iter__(self):
    min_len = self.minimum_word_length()
    max_len = self.maximum_word_length()
    for i in range(min_len, max_len + 1):
        yield from self.words_of_length(i)
```

Also, props to you for using `yield from`

Tagl:
@caleb531 @eliotwrobson I'm thinking I'll limit the scope of this PR to just FA since those are both regular languages. I will therefore implement for NFA as well but make another PR later on for PDA.

automata/fa/dfa.py
```diff
@@ -30,6 +30,8 @@ def __init__(self, *, states, input_symbols, transitions,
         self.initial_state = initial_state
         self.final_states = final_states.copy()
         self.allow_partial = allow_partial
+        self._word_cache = []
```
caleb531, line 33: @Tagl You should probably rebase on top of my latest commits to the `immutability` branch. I've made the DFA class immutable by overriding `__setattr__` and `__delattr__`—this looks to be using the old `DFA.__init__` code.

I would ask that you keep the `_word_cache` scoped to the methods that need it, passing it around as a function argument as needed.

caleb531 (changes requested):
@Tagl A few minor things here and there, but I think the big request I have is related to refactoring `word_cache` to be scoped to the method rather than stored as an attribute on the entire instance.

automata/fa/dfa.py
Tagl: The reason for the while loop is that `limit` can be `float('inf')`

automata/fa/dfa.py
caleb531: @Tagl That's a fair point. And the fact that this is a private (and volatile, i.e. caching) property means the fact that it changes isn't such a big deal.

Then my recommendation would be: when you rebase off the latest `immutability` branch, feel free to create the cache initially using the `super().__init__` call, then because you should not need to reassign to `_word_cache`, you should be able to push to it freely.

This will probably break my existing `test_dfa_immutable_dict` test because in that test, I am assuming that all attrs on the DFA's `__dict__` are hashable. But I can probably adjust that, or maybe expose a separate method for retrieving the `__dict__` with internal properties excluded.

@eliotwrobson What do you think?

eliotwrobson: @caleb531 I think that course of action sounds reasonable. @Tagl Is right that these really should be stored along with the lifetime of the object, since having to recompute them is expensive, and it really should be done as-needed (the way that it is done here). I think testing for immutability excluding internal properties is a good plan.

caleb531:
@Tagl If it's alright, I'm going to rebase your `enumeration` branch on top of the latest `immutability` branch.

caleb531:
@Tagl Please don't push any more commits to `enumeration` for the time being. I am still rebasing.

automata/fa/dfa.py
caleb531: @Tagl I see. `range()` can't receive `float` values, so I understand the need to use a `while` now.

automata/fa/dfa.py
caleb531: @eliotwrobson Agreed. I've also added a new `Automaton.attributes()` method to this commit that basically returns `__dict__` but with only the public attributes. I might rename this later to be more semantically-clear, but not until after this PR is merged.

caleb531:
@Tagl Okay, rebase is complete. I'm all done working on the `enumeration` branch.

The last thing I would really like to see in this PR is documentation for these new public methods in the README. I realized we'll all need to make sure to keep the README up to date with all these PRs being merged, otherwise it'll become a major burden on me to put that all together. I would ask if we could update the README incrementally (as needed) with each PR as it has been in the past.

@eliotwrobson would you have time to send me a separate PR with README updates for your recently-added public methods? (any README updates for #82 can just be part of that PR). That would be a great help to me.

eliotwrobson:
@caleb531 Sure, I'll look through the last few PRs I've added and update the README on #82. There's a slight change on the public methods for #77 (adding a couple of parameters), so I'll make the update for that one localized there.

tests/test_dfa.py
Tagl: I added a test for iter that does this

automata/fa/dfa.py
```diff
@@ -606,6 +718,11 @@ def from_finite_language(cls, language, input_symbols):
         Chapter 10
         """
 
+        if len(language) == 0:
```
eliotwrobson, line 721: Can you have this throw an exception instead? I want to write a class method that returns the empty DFA later.

Similarly, I think this algorithm doesn't work with the empty string, so it should throw an exception in that case.

Tagl: One of my tests for len was using the function to create the empty language, which is finite, and therefore should be possible to create with this function in my opinion. I don't think we should be raising an exception in that case.

The algorithm does work with the empty string, at least what I've tried so far seems to work.

Tagl: However I do agree with adding a specific function for creating the empty language and this function can then defer to that function for this edge case.

automata/fa/dfa.py
```diff
@@ -591,12 +707,16 @@ def isfinite(self):
...
-            nx.find_cycle(G.subgraph(important_nodes))
-            return False
-        except nx.exception.NetworkXNoCycle:
-            return True
+            return nx.dag_longest_path_length(subgraph)
+        except nx.exception.NetworkXUnfeasible:
+            return float('inf')
+
+    @staticmethod
+    def _stringify_states_unsorted(states):
```
eliotwrobson, line 717: I think this is here by accident? We removed this function a little while ago.

automata/fa/dfa.py
```diff
@@ -581,6 +605,98 @@ def isfinite(self):
...
+
+    def minimum_word_length(self):
+        """
+        Returns the length of the shortest word in the language represented by the DFA
+        """
+        queue = deque()
+        distances = defaultdict(lambda: float('inf'))
+        distances[self.initial_state] = 0
+        queue.append(self.initial_state)
+        while len(queue):
```
eliotwrobson, line 683:
```suggestion
        while queue:
```

automata/fa/dfa.py
eliotwrobson: Sounds good. Did you try the empty string with at least one other string? I think that might be the situation that causes problems.

automata/fa/dfa.py
Tagl: Merge accident

automata/fa/dfa.py
Tagl: Updated

automata/fa/dfa.py
Tagl: I tried the language with only the empty word and then two other finite languages that were larger but also included the empty word.

Tagl:
@caleb531 Since this contains some common bugfixes too I'll leave the PR as just DFA and make another one for NFA so you can get the bugfixes merged sooner.

automata/fa/dfa.py
caleb531: @Tagl Oh, I see. Sorry about that—that was an artifact I must've introduced during the rebase.

eliotwrobson (approved):
Looks good! I think it'll be good to merge this then integrate it with the new methods in the DFA construction.

Tagl:
@eliotwrobson I just made a pretty important change to the generation. I was under the impression that dict and set iteration followed insertion order since Python 3.6 but I was getting seemingly random orders. Since each value is guaranteed to only generate once we can use lists instead for the cache and then maintain proper lexicographical ordering.

eliotwrobson:
@Tagl sounds good. If you think using a list will affect performance, I think you could use an ordered dict instead.

automata/fa/dfa.py
```diff
@@ -581,6 +601,99 @@ def isfinite(self):
...
+        """
+        Populate word cache up to length k
+        """
+        sorted_symbols = sorted(self.input_symbols)
+        while len(self._word_cache) <= k:
+            i = len(self._word_cache)
+            self._word_cache.append(defaultdict(list))
+            level = self._word_cache[i]
+            if i == 0:
+                level.update({state: {''} for state in self.final_states})
```
caleb531, line 651: @Tagl (cc @eliotwrobson) If you updated the `defaultdict` to use a `list` instead of a `set`, shouldn't this dict comprehension also follow suit?

```python
level.update({state: [''] for state in self.final_states})
```

tests/test_dfa.py
```diff
@@ -1452,4 +1452,230 @@ def test_dfa_repr(self):
...
+            final_states={'p0', 'p1'}
+        )
+
+        generator = iter(A)
+        expected = ['',
+                    '0', '1',
+                    '00', '01', '10',
+                    '000', '001', '010', '100', '101',
+                    '0000', '0001', '0010', '0100', '0101', '1000', '1001', '1010']
+        generated_list = [next(generator) for _ in expected]
```
caleb531, line 1490: @Tagl Couldn't this be rewritten as:

```python
generated_list = list(A)
```

I think that would be more consistent with how the user will actually consume the iterable.

tests/test_dfa.py
```diff
@@ -1452,4 +1452,230 @@ def test_dfa_repr(self):
...
+                    'baa', 'babb', 'bbaa', 'bbabb', 'bbbab'}
+        dfa = DFA.from_finite_language(language, {'a', 'b'})
+        generated_set = {word for word in dfa}
+        self.assertEqual(generated_set, language)
+
+    def test_iter_infinite(self):
+        """
+        Test that language that avoids the pattern '11' generates the correct values in correct order
+        """
+        A = DFA(
```
caleb531, line 1472: @Tagl Could you please rename `A` to something like `dfa` or `dfa1`? The capital letter makes me think `A` is a class rather than an instance.

automata/base/automaton.py
```diff
@@ -108,5 +112,9 @@ def __repr__(self):
         """Return a string representation of the automaton."""
         values = ', '.join(
             f'{attr_name}={self._get_repr_friendly_value(attr_value)!r}'
-            for attr_name, attr_value in self.__dict__.items())
+            for attr_name, attr_value in self.attributes().items())
         return f'{self.__class__.__qualname__}({values})'
+
+    def __contains__(self, other):
```
caleb531, line 118: @Tagl Since `other` does not represent another automaton instance, but rather, the input string, could you rename `other` to `input_str` here? This would be more consistent with methods like `read_input_stepwise`.

caleb531 (changes requested):
@Tagl This is looking great overall, just have a few style requests (and one question related to the `defaultdict`).

automata/fa/dfa.py
Tagl: Yes

tests/test_dfa.py
Tagl: No, it can't be done here because the iteration is infinite so I only iterate as many times as there are items in `expected`.

tests/test_dfa.py
Tagl: Will do

automata/fa/dfa.py
Tagl: Updated

tests/test_dfa.py
Tagl: Updated

automata/base/automaton.py
Tagl: Updated

tests/test_dfa.py
caleb531: @Tagl Ah yes, I'm not used to working with potentially-infinite generators 😅. Makes sense.

tests/test_dfa.py
caleb531: Thanks, although I think there's still one hanging around:

Screenshot 2022-10-27 at 1 42 56 PM

Tagl: Missed one originally, updated again.

caleb531 (approved):
@Tagl This all looks great now! Thank you so much for writing this and for making my requested changes.

PR #82, opened 2022-10-25
Add NFA.intersection()

eliotwrobson:
Adds an NFA intersection algorithm with unit tests. @caleb531 sorry to say but, the tests couldn't use the native NFA equality checking algorithm (as it was raising errors). I think there may be a bug 😬

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Automata
 
-*Copyright 2016-2022 Caleb Evans*  
+*Copyright 2016-2022 Caleb Evans*
 *Released under the MIT license*
 
 [![tests](https://github.com/caleb531/automata/actions/workflows/tests.yml/badge.svg)](https://github.com/caleb531/automata/actions/workflows/tests.yml)
@@ -460,6 +460,18 @@ new_nfa = nfa1.union(nfa2)
 
 ```python
 new_nfa = nfa1 | nfa2
+```
+
+#### NFA.intersection(self, other)
+
+Returns intersection of two NFAs
+
+```python
+new_nfa = nfa1.intersection(nfa2)
+```
+
+```python
+new_nfa = nfa1 & nfa2
 ```
 
 #### NFA.eliminate_lambda(self)
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -527,7 +527,6 @@ class NFA(fa.FA):
             final_states=new_final_states
         )
 
-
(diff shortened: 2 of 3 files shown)


automata/fa/nfa.py
```diff
@@ -81,6 +81,13 @@ def __or__(self, other):
         else:
             raise NotImplementedError
 
+    def __and__(self, other):
+        """Return the union of this NFA and another NFA."""
+        if isinstance(other, NFA):
+            return self.intersection(other)
+        else:
+            raise NotImplementedError
```
Tagl, line 89: Maybe change to `return NotImplemented`? Need to update all DFA operators then as well

Tagl (approved):
We might want to use `defaultdict` to avoid all the None checks that clutter the code but this looks good and seems like what was described in the document you sent.

automata/fa/nfa.py
eliotwrobson: Yeah, I actually thought returning that would be better, but didn't want to update tests before discussion here. @caleb531 what do you think?

caleb531: @eliotwrobson Yes, because `__and__` is a binary magic method, `return NotImplemented` should be used instead. This is the rule of thumb that I follow:

## NotImplemented

> should be returned by the binary special methods (e.g. __eq__(), __lt__(), __add__(), __rsub__(), etc.) to indicate that the operation is not implemented with respect to the other type

## NotImplementedError

> [...] In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

Source: https://stackoverflow.com/a/44575926

caleb531: @eliotwrobson I can take care of the DFA class right now. 🙂

eliotwrobson: @caleb531 Awesome! I think this PR should be good to go now.

README.md
```diff
@@ -1,6 +1,6 @@
 # Automata
 
-*Copyright 2016-2022 Caleb Evans*  
+*Copyright 2016-2022 Caleb Evans*
```
caleb531, line 3: @eliotwrobson Just a note for the future: your editor seems to keep removing the two trailing spaces that is valid in Markdown to insert a line break. Make sure you have the editorconfig extension installed in your editor.

caleb531 (approved):

PR #83, opened 2022-10-26
Add DFA construction helpers

Tagl:
Added the following construction helpers:
- `universal_language`
- `empty_language`
- `from_substring`
- `from_subsequence`
- `from_prefix`
- `from_suffix`
- `of_length`
- `count_mod`
- `nth_from_end`
- `nth_from_start`

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ minimal_symmetric_difference_dfa = dfa ^ other_dfa
 symmetric_difference_dfa = dfa.symmetric_difference(other_dfa, minify=False)
 ```
 
-### DFA.issubset(self, other_dfa)
+#### DFA.issubset(self, other_dfa)
 
 Given two DFAs which accept the languages A and B respectively,
 returns True of the A is a subset of B, False otherwise.
@@ -280,7 +280,7 @@ dfa <= other_dfa
 dfa.issubset(other_dfa)
 ```
 
-### DFA.issuperset(self, other_dfa)
+#### DFA.issuperset(self, other_dfa)
 
 Given two DFAs which accept the languages A and B respectively,
 returns True of the A is a superset of B, False otherwise.
@@ -290,7 +290,7 @@ dfa >= other_dfa
 dfa.issuperset(other_dfa)
 ```
 
-### DFA.isdisjoint(self, other_dfa)
+#### DFA.isdisjoint(self, other_dfa)
 
 Given two DFAs which accept the languages A and B respectively,
 returns True of A and B are disjoint, False otherwise.
@@ -299,7 +299,7 @@ returns True of A and B are disjoint, False otherwise.
 dfa.isdisjoint(other_dfa)
 ```
 
-### DFA.isempty(self)
+#### DFA.isempty(self)
 
 Returns `True` if the DFA does not accept any inputs, False otherwise.
 
@@ -307,7 +307,7 @@ Returns `True` if the DFA does not accept any inputs, False otherwise.
 dfa.isempty()
 ```
 
-### DFA.isfinite(self)
+#### DFA.isfinite(self)
 
 Returns `True` if the DFA accepts a finite language, False otherwise.
 
@@ -315,7 +315,7 @@ Returns `True` if the DFA accepts a finite language, False otherwise.
 dfa.isfinite()
 ```
 
-### DFA.minimum_word_length(self)
+#### DFA.minimum_word_length(self)
 
 Returns the length of the shortest word in the language represented by the DFA.
 
@@ -323,7 +323,7 @@ Returns the length of the shortest word in the language represented by the DFA.
 dfa.minimum_word_length()
 ```
 
-### DFA.maximum_word_length(self)
+#### DFA.maximum_word_length(self)
 
 Returns the length of the longest word in the language represented by the DFA.
 In the case of infinite languages, `float('inf')` is returned.
@@ -332,7 +332,7 @@ In the case of infinite languages, `float('inf')` is returned.
 dfa.maximum_word_length()
 ```
 
-### DFA.count_words_of_length(self, k)
+#### DFA.count_words_of_length(self, k)
 
 Counts words of length `k` accepted by the DFA.
 
@@ -340,7 +340,7 @@ Counts words of length `k` accepted by the DFA.
 dfa.count_words_of_length(3)
 ```
 
-### DFA.words_of_length(self, k)
+#### DFA.words_of_length(self, k)
 
 Generates words of length `k` accepted by the DFA.
 
@@ -350,6 +350,7 @@ for word in dfa.words_of_length(3):
 ```
 
 You can also iterate through all words accepted by the DFA.
+Be aware that the generator may be infinite.
 
 ```python
 for word in dfa:
@@ -358,7 +359,7 @@ for word in dfa:
     print(word)
 ```
 
-### DFA.cardinality(self)
+#### DFA.cardinality(self)
 
 Returns the cardinality of the language represented by the DFA.
 Note that `len(dfa)` raises a `ValueError` for infinite languages, whereas `DFA.cardinality` will return `float('inf')`.
@@ -368,35 +369,83 @@ dfa.cardinality()
 len(dfa)
 ```
 
-#### DFA.contains_substring(cls, input_symbols, substring)
+#### DFA.from_prefix(cls, input_symbols, prefix, contains=True)
+
+Directly computes the minimal DFA recognizing strings with the
+given prefix.
+If `contains` is set to `False` then the complement is constructed instead.
+
+```python
+contains_prefix_nano = DFA.from_prefix({'a', 'n', 'o', 'b'}, 'nano')
+avoids_prefix_nano = DFA.from_prefix({'a', 'n', 'o', 'b'}, 'nano', contains=False)
+```
+
+#### DFA.from_suffix(cls, input_symbols, suffix, contains=True)
+
+Directly computes the minimal DFA recognizing strings with the
+given prefix.
+If `contains` is set to `False` then the complement is constructed instead.
+
+```python
+contains_suffix_nano = DFA.from_suffix({'a', 'n', 'o', 'b'}, 'nano')
+avoids_suffix_nano = DFA.from_suffix({'a', 'n', 'o', 'b'}, 'nano', contains=False)
+```
+
+#### DFA.from_substring(cls, input_symbols, substring, contains=True, must_be_suffix=False)
 
 Directly computes the minimal DFA recognizing strings containing the
 given substring.
+If `contains` is set to `False` then the complement is constructed instead.
+If `must_be_suffix` is set to `True`, then the substring must be a suffix instead.
 
 ```python
-dfa = DFA.contains_substring({'a', 'n', 'o', 'b'}, 'nano')
+contains_substring_nano = DFA.contains_substring({'a', 'n', 'o', 'b'}, 'nano')
+avoids_substring_nano = DFA.contains_substring({'a', 'n', 'o', 'b'}, 'nano', contains=False)
 ```
 
-#### DFA.contains_subsequence(cls, input_symbols, subsequence)
+#### DFA.from_subsequence(cls, input_symbols, subsequence, contains=True)
 
-Creates a DFA which accepts all words which contain a specific subsequence of symbols
+Directly computes the minimal DFA recognizing strings containing the
+given subsequence.
+If `contains` is set to `False` then the complement is constructed instead.
 
 ```python
-dfa = DFA.contains_subsequence({'a', 'b', 'c', 'd'}, 'dcba')
+contains_substring_dcba = DFA.contains_subsequence({'a', 'b', 'c', 'd'}, 'dcba')
+avoids_substring_dcba = DFA.contains_subsequence({'a', 'b', 'c', 'd'}, 'dcba', contains=False)
 ```
 
-#### DFA.of_length(cls, input_symbols, min_length=0, max_length=float('inf'))
+#### DFA.of_length(cls, input_symbols, min_length=0, max_length=None)
 
-Creates a DFA which accepts all words whose length is between `min_length` and `max_length`, inclusive.
-To allow infinitely long words the value `float('inf')` can be passed in for `max_length`.
+Directly computes the minimal DFA which accepts all words whose length is between `min_length` and `max_length`, inclusive.
+To allow infinitely long words the value `None` can be passed in for `max_length`.
 
 ```python
-dfa = DFA.of_length({'0', '1'}, 4, float('inf'))
+dfa = DFA.of_length({'0', '1'}, min_length=4)
+dfa = DFA.of_length({'0', '1'}, min_length=4, max_length=8)
+```
+
+#### DFA.count_mod(cls, input_symbol, k, remainders=None, symbols_to_count=None)
+
+Directly computes a DFA that counts given symbols and accepts all strings where
+the remainder of division by `k` is in the set of `remainders` given.
+The default value of `remainders` is `{0}` and all symbols are counted by default.
+
+```python
+even_length_strings = DFA.count_mod({'0', '1'}, 2)
+odd_number_of_ones = DFA.count_mod({'0', '1'}, 2, remainders={1}, symbols_to_count={'1'})
+```
+
+#### DFA.nth_from_start(cls, input_symbols, symbol, n)
+
+Directly computes the minimal DFA which accepts all words whose `n`-th character from the end is `symbol`, where `n` is a positive integer.
+
+```python
+dfa = DFA.nth_from_start({'0', '1'}, '1', 4)
 ```
 
 #### DFA.nth_from_end(cls, input_symbols, symbol, n)
 
-Creates a DFA which accepts all words whose `n`-th character from the end is `symbol`, where `n` is a positive integer.
+Directly computes the minimal DFA which accepts all words whose `n`-th character from the end is `symbol`, where `n` is a positive integer.
 
 ```python
 dfa = DFA.nth_from_end({'0', '1'}, '1', 4)
@@ -413,7 +462,7 @@ from automata.fa.nfa import NFA
 dfa = DFA.from_nfa(nfa)  # returns an equivalent DFA
 ```
 
-#### DFA.from_finite_language(self, language, input_symbols)
+#### DFA.from_finite_language(cls, input_symbols, language)
 
 Constructs the minimal DFA corresponding to the given finite language and input symbols.
 
--- a/README.md
+++ b/README.md
@@ -368,6 +368,31 @@ dfa.cardinality()
 len(dfa)
 ```
 
+#### DFA.contains_subsequence(cls, subsequence, input_symbols)
+
+Creates a DFA which accepts all words which contain a specific subsequence of symbols
+
+```python
+dfa = DFA.contains_subsequence('dcba', {'a', 'b', 'c', 'd'})
+```
+
+#### DFA.of_length(cls, subsequence, input_symbols)
+
+Creates a DFA which accepts all words whose length is between `min_length` and `max_length`, inclusive.
+To allow infinitely long words the value `float('inf')` can be passed in for `max_length`.
+
+```python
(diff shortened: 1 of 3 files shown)


eliotwrobson:
@Tagl I actually have the `contains_substring` function written, along with some extra tests you might find useful for the `contains_subsequence`. Would you like me to post them here so you can include in this PR?

tests/test_dfa.py
```diff
@@ -1453,3 +1453,67 @@ def test_dfa_repr(self):
...
+        dfa = DFA.nth_from_end('0', 2, {'0', '1'})
+        self.assertFalse(dfa.isfinite())
+
+        dfa = DFA.nth_from_end('0', 3, {'0', '1'})
+        self.assertFalse(dfa.isfinite())
+
+        dfa = DFA.nth_from_end('1', 4, {'0', '1'})
+        self.assertFalse(dfa.isfinite())
+
+    def test_empty_language(self):
```
eliotwrobson, line 1508: Higher up in this file, there's a test that uses DeMorgan's laws along with the empty and universal languages. It would be good to use the functions you wrote in those tests. I think there may also be other tests that use the empty language (and include redundant generator code for those DFAs).

Tagl:
> @Tagl I actually have the `contains_substring` function written, along with some extra tests you might find useful for the `contains_subsequence`. Would you like me to post them here so you can include in this PR?

Absolutely, let's try to get most of these useful construction functions together into this one PR

automata/fa/dfa.py
```diff
@@ -598,6 +598,79 @@ def isfinite(self):
...
+        """
+        Creates a DFA which accepts all words whose `n`-th character from the end is `symbol`,
+        where `n` is a positive integer.
+        """
+        # TODO: special case for len(input_symbols) == 1?
+        if n < 1:
+            raise ValueError("Integer must be positive")
+        if symbol not in input_symbols:
+            raise ValueError("Desired symbol is not in the set of input symbols")
+        state_count = 2**n
```
eliotwrobson, line 662: Doesn't this only work if there are two input symbols?

tests/test_dfa.py
Tagl: Yeah I'm thinking a lot of tests can be updated since many of the DFA we test with are either "contains this pattern" or "avoids this pattern" where the patterns are either substrings or subsequences.

automata/fa/dfa.py
```diff
@@ -598,6 +598,79 @@ def isfinite(self):
...
+        )
+
+    @classmethod
+    def of_length(cls, min_length, max_length, input_symbols):
+        """
+        Creates a DFA which accepts all words whose length is between `min_length` and `max_length`, inclusive.
+        To allow infinitely long words the value `float('inf')` can be passed in for `max_length`.
+        """
+        transitions = {}
+        length_range = range(min_length) if max_length == float('inf') else range(max_length+1)
```
eliotwrobson, line 628: I think it's better to use `None` as a sentinel value, just because using a float means passing in two different types into this function. @caleb531 what do you think?

automata/fa/dfa.py
```diff
@@ -598,6 +598,79 @@ def isfinite(self):
...
+            final_states=final_states,
+        )
+
+    @classmethod
+    def universal_language(cls, input_symbols):
+        return cls.contains_subsequence('', input_symbols)
+
+    @classmethod
+    def empty_language(cls, input_symbols):
+        return ~cls.universal_language(input_symbols)
```
eliotwrobson, line 649: Kindof a nitpick, but using `~` actually involves some unnecessary function call overhead (since this operation minimizes by default), I think it would be better just to hardcode this construction (since it's only a couple lines long and can be copied from elsewhere).

eliotwrobson:
@Tagl Per discussion, here's the substring and subsequence creation functions I had:
```python
    @classmethod
    def contains_subsequence(cls, subsequence, input_symbols):
        """
        Directly computes the minimal DFA recognizing strings containing the
        given subsequence.
        """
        transitions = {
            i: {
                symbol: i+1 if symbol == chr else i
                for symbol in input_symbols
            }
            for i, chr in enumerate(subsequence)
        }

        final_state = len(subsequence)
        transitions[final_state] = {
            symbol: final_state for symbol in input_symbols
        }

        return cls(
            states=set(transitions.keys()),
            input_symbols=input_symbols,
            transitions=transitions,
            initial_state=0,
            final_states={final_state},
        )

    @classmethod
    def contains_substring(cls, substring, input_symbols):
        """
        Directly computes the minimal DFA recognizing strings containing the
        given substring.
        """
        prefixes = [substring[:i] for i in range(len(substring))]

        transitions = {prefix: dict() for prefix in prefixes}
        transitions[substring] = {
            symbol: substring for symbol in input_symbols
        }

        for prefix in prefixes:
            prefix_dict = transitions.setdefault(prefix, dict())

            for symbol in input_symbols:
                # Look for next state after reading in the given input symbol
                possible_suffix = prefix + symbol

                # This while loop will always terminate, since the empty string is in the dict
                while possible_suffix not in transitions:
                    possible_suffix = possible_suffix[1:]

                prefix_dict[symbol] = possible_suffix

        return cls(
            states=set(transitions.keys()),
            input_symbols=input_symbols,
            transitions=transitions,
            initial_state='',
            final_states={substring},
        )
```
And here are the tests I mentioned
```python
def test_contains_substring(self):
        """Should compute the minimal DFA accepting strings with the given substring"""

        input_symbols = {'a', 'n', 'o', 'b'}

        equiv_dfa = DFA(
            states={'', 'n', 'na', 'nan', 'nano'},
            input_symbols=input_symbols,
            transitions={
                '': {'a': '', 'n': 'n', 'o': '', 'b': ''},
                'n': {'a': 'na', 'n': 'n', 'o': '', 'b': ''},
                'na': {'a': '', 'n': 'nan', 'o': '', 'b': ''},
                'nan': {'a': 'na', 'n': 'n', 'o': 'nano', 'b': ''},
                'nano': {'a': 'nano', 'n': 'nano', 'o': 'nano', 'b': 'nano'},
            },
            initial_state='',
            final_states={'nano'}
        )

        substring_dfa = DFA.contains_substring('nano', input_symbols)
        self.assertEqual(len(substring_dfa.states), len(substring_dfa.minify().states))

        self.assertEqual(len(substring_dfa.states), len(equiv_dfa.states))
        self.assertEqual(substring_dfa, equiv_dfa)

        subset_dfa = DFA.from_finite_language(['nano', 'bananano', 'nananano', 'naonano'], input_symbols)
        self.assertTrue(subset_dfa < substring_dfa)

    def test_contains_subsequence(self):
        """Should compute the minimal DFA accepting strings with the given subsequence"""

        input_symbols = {'a', 'n', 'o', 'b'}

        equiv_dfa = DFA(
            states={'', 'n', 'na', 'nan', 'nano'},
            input_symbols=input_symbols,
            transitions={
                '': {'a': '', 'n': 'n', 'o': '', 'b': ''},
                'n': {'a': 'na', 'n': 'n', 'o': 'n', 'b': 'n'},
                'na': {'a': 'na', 'n': 'nan', 'o': 'na', 'b': 'na'},
                'nan': {'a': 'nan', 'n': 'nan', 'o': 'nano', 'b': 'nan'},
                'nano': {'a': 'nano', 'n': 'nano', 'o': 'nano', 'b': 'nano'},
            },
            initial_state='',
            final_states={'nano'}
        )

        subsequence_dfa = DFA.contains_subsequence('nano', input_symbols)
        self.assertEqual(len(subsequence_dfa.states), len(subsequence_dfa.minify().states))

        self.assertEqual(len(subsequence_dfa.states), len(equiv_dfa.states))
        self.assertEqual(subsequence_dfa, equiv_dfa)

        subset_dfa = DFA.from_finite_language(['naooono', 'bananano', 'onbaonbo', 'ooonano'], input_symbols)
        self.assertTrue(subset_dfa < subsequence_dfa)

        substring_dfa = DFA.contains_substring('nano', input_symbols)
        self.assertTrue(substring_dfa < subsequence_dfa)
```

automata/fa/dfa.py
Tagl: You have the symbol `symbol` which is the one you are looking for to be the start of your $n$ character suffix.
All other symbols belong to the same equivalence class of *not what you're looking for*. Therefore the construction looks something like this:
image

Here we are looking for `x` and can treat all other characters as `?`

eliotwrobson: Ahhhh I see now, thanks for the clarification 😃

automata/fa/dfa.py
Tagl: On that note, I think we should make input_symbols the first parameter in these construction methods so we can make easier use of default parameters.

automata/fa/dfa.py
Tagl: Updated

automata/fa/dfa.py
Tagl: We might want to leave a comment in the code regarding this as it is not immediately obvious.

tests/test_dfa.py
Tagl: I've updated some of the old tests. I didn't want to change anything to do with equality since in those it's best to have some manual constructions.

automata/fa/dfa.py
Tagl: Updated the ordering so now `input_symbols` is the first parameter

eliotwrobson:
@Tagl Containing multiple copies of a substring might be interesting? Also possibly counting the length / number of symbols mod k (where k is an input parameter). My ideas are starting to get a little contrived. Though the new functions so far look awesome.

automata/fa/dfa.py
```diff
@@ -710,17 +710,159 @@ def maximum_word_length(self):
...
+        return cls(
+            states=set(transitions.keys()),
+            input_symbols=input_symbols,
+            transitions=transitions,
+            initial_state='',
+            final_states={substring},
+        )
+
+    @classmethod
+    def contains_subsequence(cls, input_symbols, subsequence):
```
caleb531, line 747: @Tagl Per my comment above, can we rename this method to `from_subsequence`?

automata/fa/dfa.py
```diff
@@ -710,17 +710,159 @@ def maximum_word_length(self):
             return float('inf')
 
     @classmethod
-    def from_finite_language(cls, language, input_symbols):
+    def contains_substring(cls, input_symbols, substring):
```
caleb531, line 713: @Tagl Can we rename this method to `from_substring`? That would map better nicely to the Python convention used in this library (e.g. `from_dfa`, `from_nfa`, etc.)

automata/fa/dfa.py
```diff
@@ -710,17 +710,159 @@ def maximum_word_length(self):
...
+        return cls(
+            states=set(transitions.keys()),
+            input_symbols=input_symbols,
+            transitions=transitions,
+            initial_state=0,
+            final_states={len(subsequence)},
+        )
+
+    @classmethod
+    def of_length(cls, input_symbols, min_length=0, max_length=float('inf')):
```
caleb531, line 767: Wondering if keyword-only args would be good to use here, at least for `min_length` and `max_length. I think it would make your README example a bit cleaner:

```python
def of_length(cls, input_symbols, *, min_length=0, max_length=float('inf')):
    ...
```

```python
dfa = DFA.of_length({'0', '1'}, min_length=4, max_length=float('inf'))
```

@eliotwrobson, what do you think?

automata/fa/dfa.py
caleb531: @Tagl Agreed about adding a code comment (my future self will probably find it helpful).

README.md
```diff
@@ -368,6 +368,40 @@ dfa.cardinality()
...
+Creates a DFA which accepts all words whose length is between `min_length` and `max_length`, inclusive.
+To allow infinitely long words the value `float('inf')` can be passed in for `max_length`.
+
+```python
+dfa = DFA.of_length({'0', '1'}, 4, float('inf'))
+```
+
+#### DFA.nth_from_end(cls, input_symbols, symbol, n)
+
+Creates a DFA which accepts all words whose `n`-th character from the end is `symbol`, where `n` is a positive integer.
```
caleb531, line 399: @Tagl Please add a note to the documentation here about the `ValueError` that's raised if the `symbol` does not exist within `input_symbols`.

tests/test_dfa.py
```diff
@@ -683,23 +683,13 @@ def test_isdisjoint(self):
...
-                'q0': {'0': 'q0', '1': 'q1'},
-                'q1': {'0': 'q1', '1': 'q2'},
-                'q2': {'0': 'q2', '1': 'q3'},
-                'q3': {'0': 'q3', '1': 'q3'}
-            },
-            initial_state='q0',
-            final_states={'q3'}
-        )
+        input_symbols = {'0', '1'}
+        A = DFA.contains_subsequence(input_symbols, '111')
```
caleb531, line 687: @Tagl Similar with the past PR, please use lowercase variable names instead of `A`, `B`, `C`, etc.

caleb531 (changes requested):
@Tagl Looks good so far—left a few requests for you to review.

automata/fa/dfa.py
eliotwrobson: Sounds good, in keeping with the conventions of other functions. I think this got mentioned before, but I think that `None` is a better default sentinel value than `float('inf')`.

README.md
eliotwrobson: Isn't there a specialized `InvaildSymbol` exception somewhere?

caleb531: @eliotwrobson @Tagl Ah yes, `InvalidSymbolError`. Good call.

Tagl:
@eliotwrobson @caleb531 `from_prefix` and `from_suffix` are probably quite commonly useful too.

automata/fa/dfa.py
Tagl: We should be able to pass whatever is returned by the word length functions in to this function, in my opinion. The default sentinel value can be None but it must also support float('inf') if the `maximum_word_length` function returns float('inf').
If we switch to None then we still are passing in two different types.

automata/fa/dfa.py
Tagl: I can do that and with that naming change I would like to propose an additional optional parameter which defines whether we seek containment or avoidance.
Construction remains the same but final states will be different. This allows the user to avoid the overhead of using the complement function after construction.
My main use case for example has a lot to do with containment/avoidance of subsequences.

automata/fa/dfa.py
Tagl: Comment added

automata/fa/dfa.py
Tagl: Updated to keyword arguments

README.md
Tagl: Updated

tests/test_dfa.py
Tagl: Everything I touched in this PR has been updated

automata/fa/dfa.py
Tagl: Updated, see further discussion on other comment

automata/fa/dfa.py
eliotwrobson: Right, but I think `None` is known to be a sentinel value (enough that Python has native support for it with the `Optional` type), while the `float('inf')` is requiring the user to passing _one specific value_ of the `float` type (they can't put in `3.5`, for example). I think it's actually worth considering whether we should remove all of the `float('inf')` return types altogether for this. Ultimately it's up to @caleb531, since is more about how the library is organized rather than any serious algorithmic considerations.

caleb531: I would agree with @eliotwrobson that `None` is preferable. It is a Python convention, and semantically, it implies that "there is no maximum length" rather than "the maximum length is infinity" (the former of which makes more sense to me).

Plus, it is unnatural to me to represent a length value with a float, per @eliotwrobson's other point. Sure, we could add a library to represent infinity as an integer, but that seems a bit overkill to me, and erodes my first argument above.

Tagl: I'm fine with using `None`, there's just pros and cons with each approach.
`None` also semantically makes sense for the empty language, but there it means something completely different.
I think leaving that case as an exception though makes more sense since it truly is an exception.

I don't really like using floating points at all when I can avoid it. The reason I chose it originally was simplicity, you can compare using the `<` operator and everything will work the same without special casing.

I think we should also avoid adding dependencies unless it is heavily relied upon or adds great value. If we wanted to go that way, creating an integer data type with a sentinel infinity value is relatively simple and does not warrant adding a dependency in my opinion.

I will change all the logic to use `None` then.

automata/fa/dfa.py
eliotwrobson: @Tagl I think that parameter would be great. We can convert it to an internal helper if later on we feel like we want different user-facing function names.

caleb531:
@Tagl Just a heads up—I was refactoring the assertions in `test_dfa` and it seems to have caused some conflicts on this PR. You'll probably want to rebase on top of the latest `develop`—let me know if I can help with that.

Apologies for the inconvenience.

Tagl:
> @Tagl Just a heads up—I was refactoring the assertions in `test_dfa` and it seems to have caused some conflicts on this PR. You'll probably want to rebase on top of the latest `develop`—let me know if I can help with that.
> 
> Apologies for the inconvenience.

But I already refactored those tests 😢

Which version do you prefer? I find naming the automata something other than `dfa` when we are working with multiple together is easier to read.

automata/fa/dfa.py
Tagl: > On a related note, do you feel like having a Match object would be useful to you? Something that takes in an FSM (DFA or NFA) and uses it to search for all matches in a given input string.

No, not particularly useful for me so far.

caleb531:
@Tagl I'm sorry... I will make it up to you by taking care of the rebase.

That refactor I made was mostly to refactor the uppercase single-letter names like `A` and `B. I hear what you're saying, and I have no qualms with instance names like `at_least_four_ones`.

caleb531:
@Tagl I got frustrated with the rebase, so I decided to revert my test_dfa commit and not rebase your branch at all. I'll tweak test_dfa after this branch gets merged. 🙂

automata/fa/dfa.py
Tagl: Now my only concern is the `cardinality` function. I feel like `None` is not the proper value to return so I'm inclined to have it raise a `ValueError`, just as `len` must do.

Tagl: Updated

automata/fa/dfa.py
Tagl: @eliotwrobson @caleb531 I just pushed an optimization of this function. I was able to pass in strings of length `>100000` after the optimization with a much faster runtime than before.

Example repeated pattern `abcd`
| size of string | runtime in seconds |
| ----:| ----------------------:|
| 4 | 0.00027 |
| 40 | 0.00023 |
| 400 | 0.00234 |
| 4000 | 0.02422 |
| 40000 | 0.19565 |
| 400000 | 1.93628 |

Tagl:
> @Tagl Containing multiple copies of a substring might be interesting? Also possibly counting the length / number of symbols mod k (where k is an input parameter). My ideas are starting to get a little contrived. Though the new functions so far look awesome.

Length or symbol count modulo $k$ could be useful so I will add that but I think that will be the last thing I put in this PR and then I'll update the documentation.

Tagl:
@caleb531 @eliotwrobson Should be all ready now for review.

README.md
```diff
@@ -270,7 +270,7 @@ minimal_symmetric_difference_dfa = dfa ^ other_dfa
 symmetric_difference_dfa = dfa.symmetric_difference(other_dfa, minify=False)
 ```
 
-### DFA.issubset(self, other_dfa)
+#### DFA.issubset(self, other_dfa)
```
caleb531, line 273: @Tagl Didn't even realize the hierarchy was incorrect. Thanks for fixing this. 🙂

Tagl: You're most welcome

caleb531 (approved):
@Tagl Looks great! Thank you for this.

eliotwrobson (approved):
Looks really awesome, thank you for all the hard work @Tagl !

PR #84, opened 2022-10-26
Cleaned up regex subset algorithm

eliotwrobson:
@caleb531 Uses a cleaner regex subset algorithm (using NFA library calls to avoid DFA conversion). It might be a bad idea to include this as an NFA method directly (since it creates a new intermediate NFA), but this recipe should probably go in the documentation.


caleb531 (approved):
@eliotwrobson I very much like the look of this. Approved, and thank you!

PR #85, opened 2022-10-27
NFA extra test

eliotwrobson:
@caleb531 Adds some extra tests for regex parsing to NFAs and style in DFA minimization.

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -123,7 +123,7 @@ class NFA(fa.FA):
     @classmethod
     def from_regex(cls, regex):
         """Initialize this NFA as one equivalent to the given regular expression"""
-        input_symbols = set(regex) - {'*', '|', '(', ')', '?'}
+        input_symbols = set(regex) - {'*', '|', '(', ')', '?', ' ', '\t'}
         nfa_builder = parse_regex(regex)
 
         return cls(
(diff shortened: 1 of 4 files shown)


automata/fa/nfa.py
```diff
@@ -123,7 +123,7 @@ def _from_symbol(cls, symbol, input_symbols=None):
     @classmethod
     def from_regex(cls, regex):
         """Initialize this NFA as one equivalent to the given regular expression"""
-        input_symbols = set(regex) - {'*', '|', '(', ')', '?'}
+        input_symbols = set(regex) - {'*', '|', '(', ')', '?', ' ', '\t'}
```
caleb531, line 126: @eliotwrobson What is the purpose of the extra regex symbols?

eliotwrobson: Allows regex to ignore whitespace. The lexer is already configured to skip whitespace by default, without this change regexes will have whitespace passed in as input symbols.

caleb531: @eliotwrobson Is whitespace not significant in a automata theory regex (like it is in Perl regex)?

eliotwrobson: The way it's configured now it's just ignored. You can't use whitespace as implicit parentheses. 

This behavior can actually be changed but that's a little beyond the scope of this PR

caleb531: @eliotwrobson Makes sense. This looks good, then.

caleb531 (approved):

PR #86, opened 2022-10-27
NFA equality bugfix

eliotwrobson:
@caleb531 Fixes a bug in NFA equality and switches other tests to use that algorithm (some of these were failing before the bugfix).

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -596,7 +596,7 @@ class NFA(fa.FA):
             # If at least one of the current states is a final state, the
             # condition should satisfy
             return any(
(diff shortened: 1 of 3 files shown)


automata/fa/nfa.py
```diff
@@ -596,7 +596,7 @@ def is_final_state(states_pair):
             # If at least one of the current states is a final state, the
             # condition should satisfy
             return any(
-                nfa.final_states - nfa._lambda_closures[state]
+                len(nfa.final_states & nfa._lambda_closures[state]) > 0
```
caleb531, line 599: @eliotwrobson Oh duh, totally makes sense. 🤦🏻‍♂️

PR #89, opened 2022-10-28
Implement __slots__ everywhere

caleb531:
@eliotwrobson Posting this PR with my `__slots__` implementation to get your feedback, when you have time.

### DIFF
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -78,15 +78,16 @@ class Automaton(metaclass=abc.ABCMeta):
                 'final states are not valid ({})'.format(
                     ', '.join(str(state) for state in invalid_states)))
 
-    def attributes(self):
+    @property
+    def input_parameters(self):
         """Return the public attributes for this automaton."""
-        return {key: value
-                for key, value in vars(self).items()
-                if not key.startswith('_')}
+        return {attr_name: getattr(self, attr_name)
+                for attr_name in self.__slots__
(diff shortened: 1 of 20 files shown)


automata/base/automaton.py
```diff
@@ -78,15 +78,16 @@ def _validate_final_states(self):
                 'final states are not valid ({})'.format(
                     ', '.join(str(state) for state in invalid_states)))
 
-    def attributes(self):
+    @property
```
eliotwrobson, line 81: Isn't there a way to disable the setter if you're declaring things with the property attribute?

eliotwrobson (commented):
One comment but looks pretty alright otherwise.

automata/base/automaton.py
caleb531: @eliotwrobson From what I'm reading, `@property` is read-only by default. I'd have to declare `@input_parameters.setter` on a new method to enable the setter. But I also tested in the REPL, and our `__setattr__` override kicks in anyway.

```
>>> dfa.input_parameters = ''
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/caleb/Repositories/personal/automata/automata/base/automaton.py", line 30, in __setattr__
    raise AttributeError(f'This {type(self).__name__} is immutable')
AttributeError: This DFA is immutable
```

Sources:
https://www.programiz.com/python-programming/property
https://docs.python.org/3/library/functions.html

PR #90, opened 2022-10-29
Levenshtein NFA construction

eliotwrobson:
Adds construction of Levenshtein NFA (used to recognize words within a fixed levenshtein distance of a query word). Also optimizes the BFS in the `from_nfa` method. @caleb531


automata/fa/nfa.py
```diff
@@ -631,3 +631,50 @@ def transition(states_pair, symbol):
                     pair_stack.append((r_1, r_2))
 
         return True
+
+    @classmethod
+    def levenshtein(cls, input_symbols, query_string, D):
+        """Code adapted from http://blog.notdot.net/2010/07/Damn-Cool-Algorithms-Levenshtein-Automata """
```
Tagl, line 637: Have the docstring also explain what is constructed and how the parameters affect the construction.
I think edit distance is the simplest term for it.

Tagl (approved):
Good catch on the BFS optimization. This is actually a very common implementation mistake for BFS, made it many times myself.

automata/fa/nfa.py
```diff
@@ -634,3 +634,56 @@ def transition(states_pair, symbol):
                     pair_stack.append((r_1, r_2))
 
         return True
+
+    @classmethod
+    def levenshtein(cls, input_symbols, query_string, D):
```
caleb531, line 639: @eliotwrobson It may be the conventional variable name for this algorithm, but I instead would prefer a snake_case variant like `distance`.

caleb531: @eliotwrobson Also, can we rename `query_string` to `input_str` if that semantically makes sense?

Finally, What do you think about keyword-only arguments for everything after (not including) `input_symbols`?

caleb531 (changes requested):
@eliotwrobson Looks great overall; just have a mix of a few requests and questions.

automata/fa/nfa.py
eliotwrobson: I think that keyword-only arguments have only been used for input parameters that have default values set (and these don't have any sensible defaults). Though I think that the other renames are good.

eliotwrobson:
Got a chance to play around with this some, and it's fast enough for some fun applications! I was able to build a DFA for all the words in a reasonably sized English dictionary, intersect them with this NFA, and get all the resulting words.

automata/fa/nfa.py
eliotwrobson: @caleb531 actually, I don't think that renaming `query_string` to `input_str` is a great idea, since the second name is used to refer to the input given to the NFA (whereas this is the string that the construction is based on).

automata/base/automaton.py
```diff
@@ -44,10 +44,10 @@ def read_input(self, input_str):
 
         Return the automaton's final configuration if this string is valid.
         """
+
         # "Fast-forward" generator to get its final value
-        for config in self.read_input_stepwise(input_str):
-            pass
-        return config
+        *_, last_config = self.read_input_stepwise(input_str)
```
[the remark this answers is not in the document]
eliotwrobson: Ah that makes sense, will flip back.

automata/fa/nfa.py
Tagl: Perhaps `reference_string` could work.
Default value could be either `0` or `1` but I feel like it should always be specified so there should not be any default.
I think `max_distance` or `max_edit_distance` would be clearer than just `distance`.

eliotwrobson: Made the variable name changes you suggested @Tagl . Also I agree, the user should really specify the distance.

eliotwrobson:
@caleb531 This one is ready to go, in working on the other PRs I couldn't think of any changes I could make to this one in terms of features over the last couple of days (the newest commit was fixing a style mistake).

README.md
```diff
@@ -622,6 +622,28 @@ Removes epsilon transitions from the NFA which recognizes the same language.
 ```python
 nfa1.eliminate_lambda()
 ```
+#### NFA.edit_distance(cls, input_symbols, reference_string, max_edit_distance,
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson On second thought, I can take care of this after merge. It's such a small change, and I'd love to get this PR merged.

automata/fa/nfa.py
```diff
@@ -634,3 +634,77 @@ def transition(states_pair, symbol):
                     pair_stack.append((r_1, r_2))
 
         return True
+
+    @classmethod
+    def edit_distance(cls, input_symbols, reference_string, max_edit_distance, *,
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson Actually might be reconsidering this one. But if I want to rename the method, I can do so anytime after this PR is merged. Thanks for all this work!

caleb531 (approved):

automata/fa/nfa.py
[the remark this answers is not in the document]
eliotwrobson: Sounds good, with all the changes that have been going in I figure there are going to have to be a few passes through to make everything consistent 😅

PR #93, opened 2022-10-30
Extended regex syntax

eliotwrobson:
@caleb531 Added intersection and kleene plus to regex syntax. Open to adding more syntax (like a wildcard, for example), but those would require modifying the interface for the current regex to NFA conversion.

EDIT: cc @Tagl not sure if you have any ideas for more regex syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1326,11 +1326,13 @@ A set of tools for working with regular languages. These can be found under
 
 A regular expression with the following operations only are supported in this library:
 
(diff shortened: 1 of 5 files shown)


Tagl:
Wildcard is the only thing I can think of at the moment

caleb531 (approved):
@eliotwrobson Looks great to me. Thank you for this!

PR #97, opened 2022-11-04
Add global configuration and allow disabling automaton validation

caleb531:
Adds a new `automata.base.config` module with the following members:

1. A `Config` class that is only used for instantiating the `global_config` instance mentioned below
2. A single `Config` instance called `global_config`, which is referenced in the `automata.base.automaton` module.

README documentation still needs to be added.

@eliotwrobson @Tagl I welcome your feedback on the code I've written so far.

### DIFF
--- a/automata/base/config.py
+++ b/automata/base/config.py
@@ -1,0 +1,9 @@
+"""Global configuration for the library and its primitives"""
+
+
+class Config():
+
+    should_validate_automata = True
+
+
+global_config = Config()
--- a/automata/base/config.py
+++ b/automata/base/config.py
@@ -1,9 +1,4 @@
 """Global configuration for the library and its primitives"""
 
 
-class Config():
-
-    should_validate_automata = True
-
-
-global_config = Config()
+should_validate_automata = True
--- a/README.md
+++ b/README.md
@@ -106,6 +106,30 @@ automaton.accepts_input(word)
 
 Returns a deep copy of the automaton according to its subtype.
 
+### Disabling automatic validation
+
+By default, all Automaton instances are checked for common inconsistencies when
+they are instantiated. If inconsistencies are found, the appropriate exception
+from `automata.base.exceptions` is raised.
+
+Because this validation can be performance-intensive for large automaton
+instances with many states/transitions, you can disable the automatic validation
+using the global configuration feature (introduced in v7):
+
+```python
+import automata.base.config as global_config
+
+global_config.should_validate_automata = False
+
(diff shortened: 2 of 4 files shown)


eliotwrobson (approved):
Seems fine overall. Possibly a little hacky, but I think for something this simple that will only be used in some situations it's fine.

caleb531:
@eliotwrobson What do you find hacky about it? 😅

eliotwrobson:
@caleb531 Just using a class variable as a global flag to control behavior. The class is effectively a singleton, and class creation behavior no longer depends only on input parameters. But like I said, for something this simple I don't think it's really a big deal.

Something like this might actually be a standard pattern in other Python packages too, just something I haven't run into before.

Tagl:
> Something like this might actually be a standard pattern in other Python packages too, just something I haven't run into before.

It's not any different than some python standard library configuration options, such as `sys.setrecursionlimit`.

automata/base/config.py
```diff
@@ -0,0 +1,9 @@
+"""Global configuration for the library and its primitives"""
+
+
+class Config():
```
Tagl, line 4: If the only configuration is the global configuration, then I'd argue there shouldn't be a class, just variables in the module. Instantiation is unnecessary here.

caleb531: @Tagl You're right. I thought I would need some sort of container object so that the module importing the config doesn't just latch onto the initial value of the setting. But that would only be a problem if I import the variable directly. I can use the module itself as said container.

```python
# doesn't work
from automata.base.config import should_validate_automata
if should_validate_automata:
    ...
```

```python
# should work fine
from automata.base import config
if config.should_validate_automata:
    ...
```

caleb531:
@eliotwrobson 

> class creation behavior no longer depends only on input parameters.

Just wanted to address this: while it is true that something outside of the input parameters can change the behavior, it is only an issue if you create an invalid automaton (which, validation or none, will produce unpredictable behavior anyway). If you always create a valid automaton, then the behavior technically does not change based on the value of this new setting.

eliotwrobson:
@caleb531 that makes sense, like I said before I don't have any objections to the way that you're doing this, just something that I haven't seen a lot (mostly due to my own inexperience 😅 ). I think it looks good!

Tagl (approved):

caleb531:
I've added the appropriate documentation to the README in this PR, so this configuration functionality should now be ready to merge.

PR #99, opened 2022-11-08
Break up API Documentation into multiple files

caleb531:
@Tagl @eliotwrobson I've finished the initial reorganization the API docs, and would love some feedback when you guys have time. I decided to go with the route of adding a `docs/` directory with a dedicated file for each section (still in Markdown)

I also decided to add the relevant navigational links to each page for easier browsing, and have updated the headings now that each section has its own file (hooray for less nesting!).


eliotwrobson (approved):
@caleb531 Haven't taken a super close look, but easy to navigate and much more manageable! I'm sure there will be revisions, but I think it's better to merge this sooner rather than later to iterate quickly.

Tagl (approved):

PR #100, opened 2022-11-08
Simplify internal DFA functions

eliotwrobson:
@caleb531 Simplifies some internal functions to the DFA class, including more efficient construction of cross product transitions. Also, switches many functions to create frozensets to avoid additional copies.


caleb531 (approved):
@eliotwrobson This looks great! Thank you!

PR #101, opened 2022-11-08
Add wildcard character to regex

eliotwrobson:
Adds wildcard `.` to the regex syntax. Also makes some regex functions more robust. Leaving as a draft until my other PR is merged.

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -9,7 +9,7 @@ from pydot import Dot, Edge, Node
 
 import automata.base.exceptions as exceptions
 import automata.fa.fa as fa
-from automata.regex.parser import parse_regex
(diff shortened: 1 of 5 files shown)


automata/fa/nfa.py
```diff
@@ -9,7 +9,7 @@
 
 import automata.base.exceptions as exceptions
 import automata.fa.fa as fa
-from automata.regex.parser import parse_regex
+from automata.regex.parser import parse_regex, RESERVED_CHARACTERS
```
caleb531, line 12: @eliotwrobson Kudos for making this a constant rather than inlining in the below code—it's probably better that way.

eliotwrobson:
@caleb531 rebased and should be good to go!

caleb531 (approved):
LGTM! 👍 Thank you!

PR #105, opened 2022-11-12
GNFA style

eliotwrobson:
Updates style of GNFA class. Should improve efficiency.

### DIFF
--- a/automata/fa/gnfa.py
+++ b/automata/fa/gnfa.py
@@ -10,7 +10,6 @@ import automata.base.exceptions as exceptions
 import automata.fa.fa as fa
 import automata.fa.nfa as nfa
 import automata.regex.regex as re
-from automata.regex.parser import RESERVED_CHARACTERS
 
 
 class GNFA(nfa.NFA):
@@ -127,7 +126,7 @@ class GNFA(nfa.NFA):
 
     def _validate_transition_invalid_symbols(self, start_state, paths):
         """Raise an error if transition symbols are invalid."""
-        check = self.input_symbols.copy() | RESERVED_CHARACTERS
+        check = self.input_symbols.copy() | {'*', '|', '(', ')', '?'}
 
         for regex in paths.values():
             if regex is not None and (set(regex) - check and regex != '' or not re._validate(regex)):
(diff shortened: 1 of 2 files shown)


automata/fa/gnfa.py
```diff
@@ -126,9 +126,9 @@ def from_nfa(cls, nfa):
 
     def _validate_transition_invalid_symbols(self, start_state, paths):
         """Raise an error if transition symbols are invalid."""
+        check = self.input_symbols.copy() | {'*', '|', '(', ')', '?'}
```
caleb531, line 129: @eliotwrobson Do we need to copy() `self.input_symbols` here? I thought union does not mutate the operands.

eliotwrobson: Ah I think you're right, no copy is necessary. Will change.

docs/fa/class-dfa.md
```diff
@@ -311,7 +311,7 @@ for word in dfa:
 ## DFA.cardinality(self)
 
 Returns the cardinality of the language represented by the DFA.
-Note that `len(dfa)` raises a `ValueError` for infinite languages, whereas `DFA.cardinality` will return `float('inf')`.
+Note that `len(dfa)` raises a `ValueError` for infinite languages, whereas `DFA.cardinality` will return `None`.
```
caleb531, line 314: @eliotwrobson So I think the docs does not match the current implementation (or the other way around 😅. I recently changed the current implementation for `DFA.cardinality` (on develop) raises an `InfiniteLanguageException` in the case of infinite languages (instead of returning `float('inf')`). Not sure what we should do about that.

```python
def cardinality(self):
    """Returns the cardinality of the language represented by the DFA."""
    try:
        i = self.minimum_word_length()
    except exceptions.EmptyLanguageException:
        return 0
    limit = self.maximum_word_length()
    if limit is None:
        raise exceptions.InfiniteLanguageException("The language represented by the DFA is infinite.")
    return sum(self.count_words_of_length(j) for j in range(i, limit+1))
```

eliotwrobson: I think either behavior is fine (since there's a way to check for the language being infinite without the exception getting raised), as long as the docs are consistent with the behavior.

caleb531: @eliotwrobson Okay, then for the sake of backwards-compatibility, I will update the docs to reflect the current (exception-based) behavior. You can revert your docs changes on this PR.

eliotwrobson: Sounds good!

caleb531:
@eliotwrobson I've updated the docs to reflect the current behavior re: cardinality, etc. I also things this PR looks great, so I will merge now.

PR #107, opened 2022-11-15
Regex simplification and bugfix

eliotwrobson:
@caleb531 Fixes a minor bug in the error message printed by the regex parser. Also significantly simplifies the construction for Kleene plus.

### DIFF
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -201,39 +201,16 @@ class NFARegexBuilder:
         """
         Apply the kleene plus operation to the NFA represented by this builder
         """
-        state_map = {
-            old_state_name: self.__get_next_state_name()
-            for old_state_name in self._transitions
-        }
-
-        second_initial_state = state_map[self._initial_state]
-
-        new_transitions = {
-            state_map[state]: {
-                chr: {state_map[end_state] for end_state in end_states}
-                for chr, end_states in path.items()
-            }
-            for state, path in self._transitions.items()
-        }
-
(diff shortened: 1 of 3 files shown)


automata/regex/parser.py
```diff
@@ -181,9 +181,10 @@ def concatenate(self, other):
 
         self._final_states = other._final_states
 
-    def kleene_star(self):
+    def kleene(self, *, is_kleene_star):
```
caleb531, line 184: @eliotwrobson Since it looks like the only thing that separates the Kleene Star operation from Kleene Plus is an extra conditional at the end, i.e.:

```python
if is_kleene_star:
    self._final_states.add(new_initial_state)
```

Given this, I almost feel it would be better to rename your `kleene()` method to `kleene_plus` without the `is_kleene_star` keyword argument, _then_ define `kleene_star` as a small wrapper around `kleene_plus` like so:

```python
def kleene_star(self):
    self.kleene_plus()
    self._final_states.add(self._initial_state)
```

If I wrote this correctly, I think it still checks out logically—just now that we can eliminate the `kleene()` middleman and just have separate `kleene_plus()` and `kleene_star()` methods.

Thoughts?

eliotwrobson: I think I slightly prefer the way I have it to lower the number of different methods, but your way is probably easier to understand for someone unfamiliar reading the code. Will change.

caleb531:
@eliotwrobson This looks great to me now. Thanks as always for the contribution!

Issue #109, opened 2022-12-02
Support for quantifiers and the shuffle operator in regular expressions


EduardoGoulart1:
Is there a specific reason why quantifiers for regular expressions are not supported? (so things like `a{2,10}`) If there is no specific reason, would you be open to a PR that implements it?

Also, what do you think of adding a special regex shuffle operator? This would be very handy to express languages like "all permutations of a,b,c". My proposal would be to use the `~` character and implement a separate parser for that

caleb531:
@EduardoGoulart1 I don't recall a specific reason, but I am certainly open to any PRs you would be willing to submit!

@eliotwrobson would quantifiers (like Eduardo is describing) fit within the theoretical regular expressions this library implements? I only ask because it seems like Perl/PCRE regex and the automata-based regex syntaxes are not the same, otherwise we would be using the stdlib's `re` module in more places.

EduardoGoulart1:
@caleb531 I think the quantifiers would be only syntactic sugar to avoid concatenating the same expression repeatedly. So `a^{n,m}` would get translated to something like `a..a_n(a_n+1)?...(a_m)?`

The shuffle operator might require some rework on the parser. We would need to construct the NFAs and compute their products explicitly.

EduardoGoulart1:
@caleb531 I think the quantifiers would be only syntactic sugar to avoid concatenating the same expression repeatedly. So `a^{n,m}` would get translated to something like `a..a_n(a_n+1)?...(a_m)?`

The shuffle operator might require some rework on the parser. We would need to construct the NFAs and compute their products explicitly.

eliotwrobson:
So I think quantifiers do fit (as they correspond to just concatenating a certain number of copies of an NFA with itself). I actually thought about adding this in before, but it causes some issues with the way that the default alphabet is set for regexes. If no objections to changing this though, I think it's very doable.

I also think that the shuffle operator is easy to do. I think this can actually be done with the same logic as the NFA `shuffle_product` function. I'm not sure about the `~` symbol, since this is usually used to denote compliment. In the literature, the Cyrillic sha ш is typically used (see https://link.springer.com/chapter/10.1007/978-3-031-19685-0_3#Sec2).

EduardoGoulart1:
@eliotwrobson but we still want the parser to be based on ASCII characters, right? I can look up for what other characters would better suit it

eliotwrobson:
@EduardoGoulart1 yes, sticking to ascii characters is best I think.

[...]

PR #112, opened 2022-12-05
Shuffle regex

eliotwrobson:
Added shuffle product to regex. @EduardoGoulart1 and @caleb531 resolves part of the issue in #109.

### DIFF
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -215,10 +215,10 @@ class NFARegexBuilder:
         self._initial_state = new_initial_state
         self._final_states.add(new_initial_state)
 
-    def shuffle(self, other):
+    def shuffle_product(self, other):
         """
         Apply the shuffle operation to the NFA represented by this builder and other.
-        No need for BFS since all states are acessible.
+        No need for BFS since all states are accessible.
         """
         new_state_name_dict = dict()
 
@@ -229,9 +229,8 @@ class NFARegexBuilder:
 
         new_transitions = dict()
 
-        for (q_a, transitions_a), (q_b, transitions_b) in product(
-            self._transitions.items(), other._transitions.items()
-        ):
+        transition_product = product(self._transitions.items(), other._transitions.items())
+        for (q_a, transitions_a), (q_b, transitions_b) in transition_product:
             state_dict = new_transitions.setdefault(get_state_name((q_a, q_b)), dict())
 
             for symbol, end_states in transitions_a.items():
@@ -281,7 +280,7 @@ class ShuffleToken(InfixOperator):
         return 1
 
     def op(self, left, right):
-        left.shuffle(right)
+        left.shuffle_product(right)
         return left
 
 
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -230,18 +230,15 @@ class NFARegexBuilder:
 
         new_transitions = dict()
 
-        for curr_state in product(self._transitions, other._transitions):
-            curr_state_name = get_state_name(curr_state)
+        for (q_a, transitions_a), (q_b, transitions_b) in product(self._transitions.items(), other._transitions.items()):
+            curr_state_name = get_state_name((q_a, q_b))
             state_dict = new_transitions.setdefault(curr_state_name, dict())
-            q_a, q_b = curr_state
 
-            transitions_a = self._transitions.get(q_a, dict())
             for symbol, end_states in transitions_a.items():
                 state_dict.setdefault(symbol, set()).update(
                     map(get_state_name, product(end_states, [q_b]))
                 )
 
-            transitions_b = other._transitions.get(q_b, dict())
             for symbol, end_states in transitions_b.items():
                 state_dict.setdefault(symbol, set()).update(
                     map(get_state_name, product([q_a], end_states))
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -228,16 +228,12 @@ class NFARegexBuilder:
 
         self._initial_state = get_state_name((self._initial_state, other._initial_state))
 
-        new_final_states = set()
         new_transitions = dict()
 
         for curr_state in product(self._transitions, other._transitions):
             curr_state_name = get_state_name(curr_state)
             state_dict = new_transitions.setdefault(curr_state_name, dict())
             q_a, q_b = curr_state
-
-            if q_a in self._final_states and q_b in other._final_states:
-                new_final_states.add(curr_state_name)
 
             transitions_a = self._transitions.get(q_a, dict())
             for symbol, end_states in transitions_a.items():
@@ -251,7 +247,7 @@ class NFARegexBuilder:
                     map(get_state_name, product([q_a], end_states))
                 )
 
-        self._final_states = new_final_states
+        self._final_states = set(map(get_state_name, product(self._final_states, other._final_states)))
         self._transitions = new_transitions
 
     @classmethod
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -10,7 +10,7 @@ from automata.regex.postfix import (InfixOperator, LeftParen, Literal,
                                     parse_postfix_tokens, tokens_to_postfix,
                                     validate_tokens)
 
-RESERVED_CHARACTERS = frozenset({'*', '|', '(', ')', '?', ' ', '\t', '&', '+', '.', '~'})
+RESERVED_CHARACTERS = frozenset({'*', '|', '(', ')', '?', ' ', '\t', '&', '+', '.', '^'})
 
 
 class NFARegexBuilder:
@@ -381,7 +381,7 @@ def get_regex_lexer(input_symbols):
     lexer.register_token(StringToken, r'[A-Za-z0-9]')
     lexer.register_token(UnionToken, r'\|')
     lexer.register_token(IntersectionToken, r'\&')
-    lexer.register_token(ShuffleToken, r'\~')
+    lexer.register_token(ShuffleToken, r'\^')
     lexer.register_token(KleeneStarToken, r'\*')
     lexer.register_token(KleenePlusToken, r'\+')
     lexer.register_token(OptionToken, r'\?')
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -131,7 +131,7 @@ class NFARegexBuilder:
             if epsilon_transitions_a is not None:
                 state_dict = new_transitions.setdefault(curr_state_name, dict())
                 state_dict.setdefault('', set()).update(
-                    get_state_name(state) for state in product(epsilon_transitions_a, [q_b])
+                    map(get_state_name, product(epsilon_transitions_a, [q_b]))
                 )
                 next_states_iterables.append(product(epsilon_transitions_a, [q_b]))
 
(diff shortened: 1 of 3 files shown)


automata/regex/parser.py
```diff
@@ -216,6 +216,39 @@ def option(self):
...
+        def get_state_name(state_name):
+            return new_state_name_dict.setdefault(state_name, self.__get_next_state_name())
+
+        self._initial_state = get_state_name((self._initial_state, other._initial_state))
+
+        new_transitions = dict()
+
+        for (q_a, transitions_a), (q_b, transitions_b) in product(
+            self._transitions.items(), other._transitions.items()
+        ):
```
caleb531, line 235: @eliotwrobson I might suggest creating a dedicated variable for the result of the `product(...)` so the `for` loop definition can fit on one line. Something like:

```python
transition_product = product(self._transitions.items(), other._transitions.items())
for (q_a, transitions_a), (q_b, transitions_b) in transition_product:
    # ...
```

automata/regex/parser.py
```diff
@@ -243,6 +276,17 @@ def op(self, left, right):
         return left
 
 
+class ShuffleToken(InfixOperator):
+    """Subclass of infix operator defining the shuffle operator."""
+
+    def get_precedence(self):
```
caleb531, line 281: @eliotwrobson Not necessary for this PR, but what do you think about making `precedence` a static property on the class, rather than a method? When I search through the codebase, most of these `get_precedence()` functions return integers.

Something like:

```python
class ShuffleToken(InfixOperator):
    """Subclass of infix operator defining the shuffle operator."""

    precedence = 1
```

But yeah, what do you think?

automata/regex/parser.py
```diff
@@ -216,6 +216,39 @@ def option(self):
         self._initial_state = new_initial_state
         self._final_states.add(new_initial_state)
 
+    def shuffle(self, other):
+        """
+        Apply the shuffle operation to the NFA represented by this builder and other.
+        No need for BFS since all states are acessible.
```
caleb531, line 222: @eliotwrobson Just a small typo here: "acessible" should be "accessible"

automata/regex/parser.py
```diff
@@ -216,6 +216,38 @@ def option(self):
...
+        """
+        new_state_name_dict = dict()
+
+        def get_state_name(state_name):
+            return new_state_name_dict.setdefault(state_name, self.__get_next_state_name())
+
+        self._initial_state = get_state_name((self._initial_state, other._initial_state))
+
+        new_transitions = dict()
```
caleb531, line 231: @eliotwrobson I see a lot of references in the codebase to `dict()` instead of `{}`. Is there a reason for this, besides to possibly distinguish it from an empty `set()`?

caleb531 (changes requested):
@eliotwrobson A few very small things, and I'd also add that this new operator is missing documentation under [`docs/regular-expressions.md`](https://github.com/caleb531/automata/blob/main/docs/regular-expressions.md). But otherwise, looks great!

Thank you as always for your thoughtful and well-written contributions.

automata/regex/parser.py
eliotwrobson: Yes, it's mainly just a habit of mine since I think it's less ambiguous than using `{}`. Fine switching styles though, if you'd like to keep that uniform through this project.

automata/regex/parser.py
```diff
@@ -10,7 +10,7 @@
                                     parse_postfix_tokens, tokens_to_postfix,
                                     validate_tokens)
 
-RESERVED_CHARACTERS = frozenset({'*', '|', '(', ')', '?', ' ', '\t', '&', '+', '.'})
+RESERVED_CHARACTERS = frozenset({'*', '|', '(', ')', '?', ' ', '\t', '&', '+', '.', '@'})
```
EduardoGoulart1, line 13: Tbh not sure about this character. '@' is quite used for user data (username or e-mail for example). I think that either we introduce it in a different version, or we enable some form of configuration for the parser...

I would vote for the latter because this operator is more exotic and from an API design perspective, it makes sense to make it easy to perform the common tasks. So if someone is using the shuffle operator, he/she is probably fine with passing around a few extra parameters

tests/test_regex.py
```diff
@@ -116,6 +116,16 @@ def test_wildcard(self):
...
 
+    def test_shuffle(self):
+        """Should correctly check shuffle"""
+
+        input_symbols = {'a', 'b', 'c', 'd'}
+
+        self.assertTrue(re.isequal('a@b', 'ab|ba', input_symbols=input_symbols))
+        self.assertTrue(re.isequal('ab@cd', 'abcd | acbd | cabd | acdb | cadb | cdab', input_symbols=input_symbols))
+        self.assertTrue(re.isequal('(a*)@(b*)@(c*)@(d*)', '.*', input_symbols=input_symbols))
+        self.assertTrue(re.isequal('ca@db', '(c@db)a | (ca@d)b', input_symbols=input_symbols))
```
EduardoGoulart1, line 127: Some extra tests `a@''`, `''@''` (should these two work?), `ab@ab`, `a*@ba`, `a@(b|c)`

EduardoGoulart1 (commented):
As in my comment, I'd prefer to not break existing regular expressions from other users

automata/regex/parser.py
eliotwrobson: You raise a good point about using the `@` symbol, although I'm not so sure about having the user pass in a character every time to be used. I think this would introduce the possibility of more confusion, along with having to pass through a bunch of options to the parser itself.

Maybe the percent sign `%` is less intrusive? Looking at regex syntax in other places, this symbol doesn't seem to be used anywhere (see [here](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions) and [here](https://www.brics.dk/automaton/doc/dk/brics/automaton/RegExp.html)). @caleb531 what are your thoughts?

caleb531: @eliotwrobson Well, because the percent sign is often used in URL escape sequences (e.g. `%20`), I would prefer to avoid it.

I personally liked @EduardoGoulart1's suggestion of `~`; even if it doesn't _look_ like Cyrillic sha, the waviness of `~` reminds me of shuffling (at least in my strange mind lol). As long as we document what the symbol is, it shouldn't matter if it doesn't look like sha.

caleb531: Also, as we add more symbols of special meaning to the regex grammar, I am thinking it's about time to add an `escape` function (much like `re.escape` in the stdlib) which properly escapes the appropriate literal characters (e.g. `someperson\+suffix1@gmail\.com`). If there's no way of escaping regex symbols currently, we should add this (preferably in a separate PR to recognize backslashes in an automaton regex).

automata/regex/parser.py
caleb531: @eliotwrobson I see. Sounds good to me.

tests/test_regex.py
caleb531: @EduardoGoulart1 @eliotwrobson I think the proper way to indicate empty string in a regex would be no characters at all between the obvious word boundaries. For example:

1. `a@`
2. `(x@)`

This is consistent with how the `|` operator is handled in PCRE regex (e.g. `a|`):

```
>>> import re
>>> re.search('(a|)', 'a')
<re.Match object; span=(0, 1), match='a'>
>>>  re.search('(a|)', '')
<re.Match object; span=(0, 0), match=''>
```

caleb531 (changes requested):
@eliotwrobson Left some feedback on the existing comment threads.

automata/regex/parser.py
eliotwrobson: @caleb531 that makes sense. I shied away from the `~` because the [brics java automata](https://www.brics.dk/automaton/doc/dk/brics/automaton/RegExp.html) library uses it for compliment, but I don't think it's a great idea to implement that operation for regexs. I do agree though that the `~` seems to invoke feelings of interleaving.

Currently, only alphanumeric characters are supported in the regex parsing here (which I'm now realizing is kindof a breaking change from v6, whoops). But I think some way of adding escape characters could be useful to people. I think it's just a matter of reconfiguring the lexer to treat characters coming after a backslash as a literal. But I think it should definitely be in a separate PR.

tests/test_regex.py
eliotwrobson: That sounds reasonable, but is a bit out of scope for this PR (requires some kinda technical changes to the parsing algorithm). Though you can indicate `a|` with the question mark optional, as `a?`.

automata/regex/parser.py
caleb531: @eliotwrobson Oh yeah, forgot about `~` as complement. In that case, I think I'd actually prefer to eliminate `~` as a candidate symbol, as well. Otherwise, it might give the wrong impression (like a negative lookahead or something). Even though complement isn't useful in a regex context, it seems like the regex operators have equivalent NFA operations as well (e.g. regex `*` maps to NFA Kleene star).

Maybe this "trying to find the right character" debate is an indication that we're pressing against the limit of what new operations should be part of the regex grammar anyway. I mean, sure, the ASCII constraint is potentially arbitrary (i.e. if we just used sha, then we wouldn't need to have this conversation). But the shuffle operator is really just syntactic sugar for something that can technically already be expressed as a series of `|`'s.

I never learned about a shuffle operation in a regex context, so I'm just trusting both you guys that it's not only practical, but coherently fits within the theory as well

I saw you _just_ pushed a commit to use `~`, but given how I feel above, I'm no longer sure that's the right operator. But what do you think about `^` as the symbol to use? Kinda has a combinatoric feel to it (because, well, `^` is often used for exponentiation).

eliotwrobson: @caleb531 I like the `^` as well. With respect to your other comment, it's actually pretty difficult to replicate the shuffle operation using just union `|` operators (as there's a combinatorial blowup in the number of terms needed), so adding support for the shuffle could actually be very convenient.

With respect to the larger vision, I actually kinda have the opposite viewpoint. A lot of other regex libraries are constrained by the fact that they use more efficient matching algorithms, so with this regex engine being oriented around NFA construction and available operations, I think it helps the library stand out to have these unique operations.

From a theory standpoint, the way to view these operations is being equivalent to operations on NFAs, and regex is just a shorthand syntax for this construction and operation. In that sense, any operation that can be performed with one or more NFAs makes sense theoretically, and I think it's cool that we have a chance to incorporate that here, while these ideas may not fit in as nicely with a more application-oriented regex implementation.

tests/test_regex.py
caleb531: @eliotwrobson Oh yes, of course. Then I think `a@''` and  `''@''` cases aren't necessary to have.

automata/regex/parser.py
eliotwrobson: @caleb531 No worries! I hope the discussion was as insightful for you as for me. And not a problem, the initial symbol change was just a find and replace, which is why I was able to push the changes so quickly 😆

eliotwrobson: Symbol has been changed to `^`!

automata/regex/parser.py
caleb531: @eliotwrobson "Up to 6 times longer"?? Eek that's good to know. I guess the StackOverflow has spoken once again!

"I Have Spoken" from The Mandalorian

caleb531 (approved):
This PR looks good to me now! Will merge presently. 🎉

automata/regex/parser.py
EduardoGoulart1: @caleb531 @eliotwrobson I can't react so quickly to PRs, so sorry if I arrive late in the discussions

I just saw your comment that we only support alphanumeric characters. That basically means that my fear of breaking existing code was nonsense because the current code does not support `@` anyways. For example, `NFA.from_regex('.*@email.com')` throws a `LexerError: Invalid character '@'`

I'm still getting into the codebase, still I should have checked for that first

PR #114, opened 2022-12-06
Resolve retain names issue

eliotwrobson:
Add retain names functionality to cross product. Resolves an inconsistency with expected behavior and should improve efficiency of minify function.

EDIT: Don't merge yet, just found a really dumb subtle bug in here. This also needs to wait until the shuffle product PR is merged, since that issue is present there as well (not a breaking bug).

### DIFF
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -23,6 +23,26 @@ def freeze_value(value):
         return tuple(freeze_value(element) for element in value)
     return value
 
+def get_renaming_function(counter):
+    """
+    A helper function that returns a renaming function to be used in the creation of
+    other automata. The parameter counter should be an itertools count.
+    This helper function will return the same distinct output taken from counter
+    for each distinct input.
+    """
+
+    new_state_name_dict = {}
+
+    def renaming_function(item):
+        nonlocal new_state_name_dict
+        if item in new_state_name_dict:
+            return new_state_name_dict[item]
+
+        return new_state_name_dict.setdefault(item, next(counter))
+
+    return renaming_function
+
+
 
 class PartitionRefinement:
     """Maintain and refine a partition of a set of items into subsets.
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -23,6 +23,7 @@ def freeze_value(value):
         return tuple(freeze_value(element) for element in value)
     return value
 
+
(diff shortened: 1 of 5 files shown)


caleb531:
@eliotwrobson I just merged the shuffle PR, so you can resume development on this one. Just switch it from Draft to Open when you're ready for me to have a look at it.

automata/base/utils.py
```diff
@@ -23,6 +23,26 @@ def freeze_value(value):
...
+    A helper function that returns a renaming function to be used in the creation of
+    other automata. The parameter counter should be an itertools count.
+    This helper function will return the same distinct output taken from counter
+    for each distinct input.
+    """
+
+    new_state_name_dict = {}
+
+    def renaming_function(item):
+        nonlocal new_state_name_dict
```
caleb531, line 37: @eliotwrobson You're not reassigning `new_state_name_dict` within this function, so the `nonlocal` declaration should be unnecessary.

eliotwrobson:
@caleb531 should be good to go now. Found a bug earlier where the counters were being called too eagerly and the numbers were going higher than needed. So I added the helper function in the utils file that handles all that logic when a counter is provided.

This actually raises the question of whether it's worth it to rename states in the NFA operations, but those don't really have huge implications for efficiency. Because the DFA `minify` function has to make copies of state names, keeping the complexity of the state names down when that function is called is actually pretty important.

eliotwrobson:
@caleb531 Oh no, I just meant that it might be worth adding `retain_names` logic to NFA constructions, but that it's not as big of a deal since the NFA operations aren't as sensitive to performance issues.

caleb531:
@eliotwrobson Well, I suppose that _could_ make sense from a consistency standpoint, since DFA operations already support a `retain_names` parameter. However, if we want to go to the full mile, we would really need a `NFA.minify()` method (and associated per-operation `minify` parameters) as well.

eliotwrobson:
@caleb531 unfortunately there are no known efficient algorithms for the NFA minimization problem (see [here](https://en.wikipedia.org/wiki/NFA_minimization)). My comment was really just an observation, I think things are fine as-is.

caleb531:
@eliotwrobson Gotcha. Well, this looks good to me! Merging now.

caleb531 (approved):

PR #116, opened 2022-12-07
Switched to default dict

eliotwrobson:
Switches utility function to use default dict.


caleb531 (approved):
@eliotwrobson Looks fine to me! Thanks for this

PR #117, opened 2022-12-08
Optimize partition function

eliotwrobson:
Optimize the partition function some more. Somewhat faster in certain test cases.


eliotwrobson:
Disregard

PR #119, opened 2022-12-15
Removed NFA._from_symbol method

eliotwrobson:
@caleb531 The NFA `_from_symbol` function was only being used as part of the old regex parsing algorithm and the functionality was a less general version of what's currently available with the DFA `from_finite_language` function. This PR removes `_from_symbol`.

As an aside, it _could_ be worth adding some NFA construction algorithms, but if we do then they should be more general than this function was.

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -106,24 +106,6 @@ class NFA(fa.FA):
                         start_state, input_symbol))
 
     @classmethod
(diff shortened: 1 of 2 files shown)


caleb531:
@eliotwrobson Makes sense to me—I agree with removing the method. Approved and (in a moment) merged!

PR #121, opened 2022-12-21
Add support for regex quantifiers

### LINKED ISSUE #109: Support for quantifiers and the shuffle operator in regular expressions

eliotwrobson:
Pull request for quantifiers, resolves #109

### DIFF
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -2,7 +2,7 @@
 """Classes and methods for parsing regexes into NFAs."""
 
 from collections import deque
-from itertools import chain, count, product, zip_longest
+from itertools import chain, count, product, repeat, zip_longest
 
 from automata.base.utils import get_renaming_function
 from automata.regex.lexer import Lexer
@@ -129,9 +129,9 @@ class NFARegexBuilder:
             if epsilon_transitions_a is not None:
                 state_dict = new_transitions.setdefault(curr_state_name, {})
                 state_dict.setdefault('', set()).update(
-                    map(get_state_name, product(epsilon_transitions_a, [q_b]))
+                    map(get_state_name, zip(epsilon_transitions_a, repeat(q_b)))
                 )
-                next_states_iterables.append(product(epsilon_transitions_a, [q_b]))
+                next_states_iterables.append(zip(epsilon_transitions_a, repeat(q_b)))
 
             # Get transition dict for states in other
             transitions_b = other._transitions.get(q_b, {})
@@ -140,9 +140,9 @@ class NFARegexBuilder:
             if epsilon_transitions_b is not None:
                 state_dict = new_transitions.setdefault(curr_state_name, {})
                 state_dict.setdefault('', set()).update(
-                    map(get_state_name, product([q_a], epsilon_transitions_b))
+                    map(get_state_name, zip(repeat(q_a), epsilon_transitions_b))
                 )
-                next_states_iterables.append(product([q_a], epsilon_transitions_b))
+                next_states_iterables.append(zip(repeat(q_a), epsilon_transitions_b))
 
             # Add all transitions moving over same input symbols
             for symbol in new_input_symbols:
@@ -232,12 +232,12 @@ class NFARegexBuilder:
 
             for symbol, end_states in transitions_a.items():
                 state_dict.setdefault(symbol, set()).update(
-                    map(get_state_name, product(end_states, [q_b]))
+                    map(get_state_name, zip(end_states, repeat(q_b)))
                 )
 
             for symbol, end_states in transitions_b.items():
                 state_dict.setdefault(symbol, set()).update(
-                    map(get_state_name, product([q_a], end_states))
+                    map(get_state_name, zip(repeat(q_a), end_states))
                 )
 
         self._final_states = set(map(get_state_name, product(self._final_states, other._final_states)))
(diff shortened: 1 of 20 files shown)


eliotwrobson:
@caleb531 Not _quite_ done yet, but an initial version of the quantifier syntax is working. The main obstacle I'm running into right now has to do with the way that the default input symbols are retrieved. The issue is that to create the factory function for the wildcard, the lexer needs to have all of the non-reserved symbols that appear in a given regex.

However, with the default method of retrieving the input symbols, numbers used in the quantifiers will get picked up even though they don't appear as literals in the given regex. Because of the difference between where the factory function is defined and called, this creates a potential sharp edge that may cause issues for some people. Do you have any thoughts on this? There are a couple of ways to get around that but I'm not quite sure which one would be best.

automata/regex/parser.py
```diff
@@ -140,9 +140,9 @@ def intersection(self, other):
             if epsilon_transitions_b is not None:
                 state_dict = new_transitions.setdefault(curr_state_name, {})
                 state_dict.setdefault('', set()).update(
-                    map(get_state_name, product([q_a], epsilon_transitions_b))
+                    map(get_state_name, zip(repeat(q_a), epsilon_transitions_b))
```
caleb531, line 143: @eliotwrobson What's the advantage of using `repeat(q_a)` over `[q_a]`?

automata/regex/regex.py
```diff
@@ -33,7 +34,7 @@ def isequal(re1, re2, *, input_symbols=None):
     nfa1 = NFA.from_regex(re1, input_symbols=input_symbols)
     nfa2 = NFA.from_regex(re2, input_symbols=input_symbols)
 
-    return nfa1 == nfa2
+    return DFA.from_nfa(nfa1) == DFA.from_nfa(nfa2)
```
caleb531: @eliotwrobson I thought we got the NFA equality algorithm working correctly. Is that not actually the case (thus leading you to compare the DFAs instead)?

eliotwrobson: Oh no, it was working fine but I was passing in the wrong NFA during testing and tried a bunch of random stuff before realizing it was my mistake 🤣 this got committed by accident. I think this change is reverted in the current version though.

automata/regex/parser.py
eliotwrobson: So I saw someone do this in an itertools video and mainly wanted to try it out, but after some timing turns out its slower... Another lesson on why it's not good to try and be too clever 😅

automata/regex/regex.py
caleb531: @eliotwrobson Gotcha. I can confirm it's reverted. Thanks for the clarity!

automata/regex/parser.py
caleb531: @eliotwrobson Well, how many iterations are those benchmarks based off of? I am curious if one solution scales better than the other.

eliotwrobson: This was 1 million. I'm actually curious now, let me try some different numbers.

eliotwrobson:
@caleb531 looking at this again, how do you feel about changing the default input symbols for creating a regex to something very general, like all alphas + numbers? I think it removes the awkwardness of having the default regex to NFA conversion be only over the symbols in present literals. If not, I think there's a workaround to the wildcard creation issue that keeps this same default, just makes the creation code a little messier.

eliotwrobson:
@caleb531 going to go ahead and flip this to open now that I had the chance to clean it up. The only controversial thing here is changing the default alphabet used for parsing regex, but I think this actually makes it easier if someone is just playing around with regexes. Performance/memory usage shouldn't be an issue since people can still manually set the input symbols.

automata/regex/parser.py
caleb531: @eliotwrobson Just checking on the status of your benchmarking, since this thread is the last unresolved conversation on this PR.

eliotwrobson: The changes discussed in this thread actually got merged in #130, but here is more detailed benchmarking:

image

Even at the smaller sizes, the difference is somewhat noticeable.

caleb531: @eliotwrobson Oh okay great! I'm satisfied then lol. Will resolve this thread and review the rest of your PR in preparation to approve/merge.

automata/regex/parser.py
```diff
@@ -181,40 +183,65 @@ def concatenate(self, other):
 
         self._final_states = other._final_states
 
-    def kleene_star(self):
+    def repeat(self, lower_bound, upper_bound):
```
caleb531, line 186: @eliotwrobson So I understand the introduction of the `repeat` method, but don't we still need to keep the `kleene_star` method lest we break API backwards-compatibility?

caleb531:
> Haven't updated docs completely, will do once this gets closer to being merged.

@eliotwrobson Just reviewed this PR again—everything looks good. Did you still have any documentation updates to make before I merge?

eliotwrobson:
@caleb531 nope! I added the docs updates to the PR already (just noting the quantifier inclusion and changing the default alphabet), so I think this is good to go. If we feel like there's more docs needed, that can be done before this goes on main.

caleb531:
@eliotwrobson Perfect! Will approve and merge now, then. Thank you for your work on this functionality!

caleb531 (approved):

Issue #36, opened 2021-07-06
Allow DFA transition function to be partial


alkkem:
When I studied Automata Theory (and in the literature), the transition function for a DFA could be partial, this is, that there could be some states which do not have an associated transition for a symbol in the alphabet.
In my particular case, I was trying to get this automata:
<img src="https://user-images.githubusercontent.com/72548193/124636808-3d951b80-de89-11eb-9ed5-fce5d1776c1d.png" width=60%>

But, as I said, some states do not have an associated transition for a symbol (take for example state _b_). This, in principle, accomplishes the DFA restrictions since there is no state such that for a symbol in the alphabet there is _more than one_ associated transition.
When trying to obtain the above DFA using the transition function:
```
{
'': {'a': 'a', 'b': 'b'},
'a': {'b': 'ab', 'a': 'aa'},
'b': {'b': 'bb'},
'aa': {'a': 'aa', 'b': 'ab'},
'bb': {'a': 'ba'},
'ab': {'b': 'bb'},
'ba': {'a': 'aa'}
}
```
I get an exception:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/guillermh/.local/lib/python3.9/site-packages/automata/fa/dfa.py", line 23, in __init__
    self.validate()
  File "/home/guillermh/.local/lib/python3.9/site-packages/automata/fa/dfa.py", line 67, in validate
    self._validate_transitions(start_state, paths)
  File "/home/guillermh/.local/lib/python3.9/site-packages/automata/fa/dfa.py", line 59, in _validate_transitions
    self._validate_transition_missing_symbols(start_state, paths)
  File "/home/guillermh/.local/lib/python3.9/site-packages/automata/fa/dfa.py", line 29, in _validate_transition_missing_symbols
    raise exceptions.MissingSymbolError(
automata.base.exceptions.MissingSymbolError: state b is missing transitions for symbol a
```
I think there is a method `_validate_transition_missing_symbols` in `automata.fa.dfa` which is the culprit of this thing.
I would make a pull request but right now I haven't got the time to learn how to implement and run tests and testing code coverage (I'm pretty novice in contributing tasks).
But I opened this issue in order to notify about the possible misconception about the formalities of DFA.
Thank you beforehand!

caleb531:
@alkkem But... but... it's not a bug, it's a feature! 😝

What you have stumble upon is actually a controversial topic:
[...]

PR #125, opened 2023-01-16
Support for pickling of finite automata

EduardoGoulart1:
The initial use-case that I stumbled upon was to serialize DFAs with pickle. I would get an error because it was calling __setattr__, which is disabled even though I was setting the `allow_mutable_automata` flag.

If we allow for mutable automata, then one should be able to modify its attributes.

```python
global_config.allow_mutable_automata = True
dfa_1 = DFA(
    states={0, 1},
    input_symbols={0, 1},
    transitions={0: {0: 1, 1: 0}, 1: {0: 1, 1: 1}},
    initial_state=0,
    final_states={1},
)

with open('dfa.pkl', 'wb') as f:
    pickle.dump(dfa_1, f)
with open('dfa.pkl', 'rb') as f:
    dfa_1_read = pickle.load(f)
assert dfa_1 == dfa_1_read
```

**Edit:** After some consideration, I decided to disable set/get attributes as this will mess with the `validate` function. Instead serialization is implemented directly with the setattribute/getattribute functions

### DIFF
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -33,11 +33,17 @@ class Automaton(metaclass=abc.ABCMeta):
 
     def __setattr__(self, name, value):
         """Set custom setattr to make class immutable."""
-        raise AttributeError(f'This {type(self).__name__} is immutable')
+        if global_config.allow_mutable_automata:
+            object.__setattr__(self, name, value)
+        else:
+            raise AttributeError(f'This {type(self).__name__} is immutable')
 
     def __delattr__(self, name):
         """Set custom delattr to make class immutable."""
-        raise AttributeError(f'This {type(self).__name__} is immutable')
+        if global_config.allow_mutable_automata:
+            object.__delattr__(self, name)
+        else:
+            raise AttributeError(f'This {type(self).__name__} is immutable')
 
     @abc.abstractmethod
     def read_input_stepwise(self, input_str):
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -33,17 +33,17 @@ class Automaton(metaclass=abc.ABCMeta):
 
     def __setattr__(self, name, value):
         """Set custom setattr to make class immutable."""
-        if global_config.allow_mutable_automata:
-            object.__setattr__(self, name, value)
-        else:
-            raise AttributeError(f'This {type(self).__name__} is immutable')
+        raise AttributeError(f'This {type(self).__name__} is immutable')
 
     def __delattr__(self, name):
         """Set custom delattr to make class immutable."""
-        if global_config.allow_mutable_automata:
-            object.__delattr__(self, name)
-        else:
-            raise AttributeError(f'This {type(self).__name__} is immutable')
+        raise AttributeError(f'This {type(self).__name__} is immutable')
+
+    def __getstate__(self):
+        return self.input_parameters
+
+    def __setstate__(self, d):
+        self.__init__(**d)
 
     @abc.abstractmethod
     def read_input_stepwise(self, input_str):
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -40,9 +40,13 @@ class Automaton(metaclass=abc.ABCMeta):
         raise AttributeError(f'This {type(self).__name__} is immutable')
 
     def __getstate__(self):
+        """Return the object's state, described by its input parameters"""
         return self.input_parameters
 
     def __setstate__(self, d):
+        """Restore the object state from its input parameters"""
+        # Notice that the default __setstate__ method won't work
+        #   because __setattr__ is disabled due to immutability
         self.__init__(**d)
 
     @abc.abstractmethod
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -13,11 +13,13 @@ class TestSerialization(test_fa.TestFA):
     """
 
     def test_serialize_dfa(self):
+        """Should convert a DFA to pickle serialization and reads it back"""
         s = pickle.dumps(self.dfa)
         dfa = pickle.loads(s)
         self.assertEqual(self.dfa, dfa)
 
     def test_serialize_nfa(self):
+        """Should convert a NFA to pickled representation and read it back"""
         s = pickle.dumps(self.nfa)
         nfa = pickle.loads(s)
         self.assertEqual(self.nfa, nfa)
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python3
+"""Tests for automata pickle serialization"""
+
+import pickle
+
+import tests.test_fa as test_fa
+
+
+class TestSerialization(test_fa.TestFA):
+    """
+    This tests verifies that the cycle FA -> .pkl -> FA works.
+    The test only applies to DFA and NFAs as other classes do not implement equality
+    """
+
+    def test_serialize_dfa(self):
+        s = pickle.dumps(self.dfa)
+        dfa = pickle.loads(s)
+        self.assertEqual(self.dfa, dfa)
+
+    def test_serialize_nfa(self):
+        s = pickle.dumps(self.nfa)
+        nfa = pickle.loads(s)
+        self.assertEqual(self.nfa, nfa)


automata/base/automaton.py
```diff
@@ -33,11 +33,17 @@ def validate(self):
 
     def __setattr__(self, name, value):
         """Set custom setattr to make class immutable."""
-        raise AttributeError(f'This {type(self).__name__} is immutable')
+        if global_config.allow_mutable_automata:
```
EduardoGoulart1, line 36: I considered calling self.validate() here to alert users if they perform an invalid operation. But I opted not to for 2 reasons:
1- Performance
2- Because loading DFAs with pickle would require to disable validation too (as pickle calls "setattr"). It could be done, but it makes it more cumbersome to use

Also, I go by the assumption that if people disable the immutability of the automata, then they should really know what they are doing

EduardoGoulart1:
@caleb531 On a side note, I feel like the behavior for `allow_partial` is underspecified. For example, the following (no transition function for one of the states) works, but it's unclear what it means:

```python
  dfa_1 = DFA(
      states={0, 1},
      input_symbols={0, 1},
      transitions={0: {0: 1, 1: 0}},
      initial_state=0,
      final_states={1},
      allow_partial=True,
  )
```
The code throws an error one tries to do `dfa_1 == dfa_1`

automata/base/automaton.py
caleb531: @EduardoGoulart1 Agreed. It would hurt performance to call `validate()` on every assignment, especially if the user needs to make multiple assignments in succession. My assumption has always been, for a mutable automaton, that `validate()` only runs under the following conditions:

1. Once in instantiation
2. Manually by the user

That is precisely why `validate()` is not `_validate()`—because it can be called by the user as needed. All that to say, yes, I agree with your assessment.

caleb531 (changes requested):
@EduardoGoulart1 Wow, great catch—these are definitely cases that need to be considered.

However, it looks like the code coverage has decreased—can you please add tests for these new conditionals? Probably under `tests/test_config.py` would be the best place to add them.

EduardoGoulart1:
> @EduardoGoulart1 Wow, great catch—these are definitely cases that need to be considered.
> 
> However, it looks like the code coverage has decreased—can you please add tests for these new conditionals? Probably under `tests/test_config.py` would be the best place to add them.

@caleb531 will do it later today. I will also to try to come up with a solution for the `allow_partial` issue

automata/base/automaton.py
EduardoGoulart1: @caleb531 I decided against having set/del attributes because the automata such as DFA class have the word cache which we don't want to mess-up with. Instead, to enable pickle, I directly use the __getstate__ and __setstate__ methods.

caleb531:
@EduardoGoulart1 So, after having given it some thought, I am actually seriously considering removing `allow_partial` altogether. It was well-intentioned, but creates a lot of complexity in the DFA class, and it becomes difficult to guarantee that the existing DFA operations will work correctly with partial DFAs (because, as you're finding, you need to test against multiple permutations of partial DFAs for _every_ operation).

@eliotwrobson When you have a moment, what are your thoughts on this? [Here is the original issue](https://github.com/caleb531/automata/issues/36) that prompted the introduction of `allow_partial`.

eliotwrobson:
I think I'm in agreement, the allow partial option is really just a minor convenience that can be removed. If we wanted to instead, we could have the constructor add an extra dump state that represents the same operation as allow partial for missing transitions, but even that may be overkill. Using two different assumptions for the internal representation of the DFA causes more problems than it fixes.

caleb531:
@eliotwrobson All good points, and agreed. I think a sufficient replacement might be something like a `DFA.from_partial` method or something like that (which would accept all the same parameters as the DFA constructor, but with the added behavior to fill in the gaps with the throwaway state.

However, that gets tricky when the user can make up any state names they want, so we would need some algorithm to determine the name of the throwaway state that doesn't conflict with any of the states they've defined. But maybe we could have `DFA.from_partial` accept a required keyword parameter that make the user supply the name of that throwaway state. Then there's no ambiguity, and it would enable the user to make the resulting DFA consistent with the existing states from their partial DFA.

Unless we do something like `retain_names=False` where we just generate a DFA with our own state names. I hope you see what I'm getting at here—what are your thoughts?

@EduardoGoulart1 please feel free to chime in as well, if you so desire.

eliotwrobson:
I think having an optional parameter for a throwaway state is fine. By default, we can just use the smallest nonnegative integer that doesn't conflict with the existing state names (I think something similar is used in some of the DFA algorithms)

eliotwrobson:
Related to this, @EduardoGoulart1 do you think there's a way to enable pickling support for immutable automatons? I think it would be nice to always have, especially if people are making large ones.

EduardoGoulart1:
> Related to this, @EduardoGoulart1 do you think there's a way to enable pickling support for immutable automatons? I think it would be nice to always have, especially if people are making large ones.

This PR should address exactly that... I changed it so that you don't need to set `allow_mutable` for pickling

EduardoGoulart1:
> @eliotwrobson All good points, and agreed. I think a sufficient replacement might be something like a `DFA.from_partial` method or something like that (which would accept all the same parameters as the DFA constructor, but with the added behavior to fill in the gaps with the throwaway state.
> 
> However, that gets tricky when the user can make up any state names they want, so we would need some algorithm to determine the name of the throwaway state that doesn't conflict with any of the states they've defined. But maybe we could have `DFA.from_partial` accept a required keyword parameter that make the user supply the name of that throwaway state. Then there's no ambiguity, and it would enable the user to make the resulting DFA consistent with the existing states from their partial DFA.
> 
> Unless we do something like `retain_names=False` where we just generate a DFA with our own state names. I hope you see what I'm getting at here—what are your thoughts?
> 
> @EduardoGoulart1 please feel free to chime in as well, if you so desire.

@caleb531 and @eliotwrobson I don't think that the library should be guided by my specific needs, but `allow_partial` is actually useful for my use case. For instance, think about a "linear" DFA with a large alphabet. Not only is it more memory efficient, but it can also be more computationally efficient e.g. for [minimization](https://www.cs.cmu.edu/~cdm/papers/Valmari12.pdf).

On the other hand, I'm not too deep into the codebase to know how much pain `allow_partial` causes us. My impression is that if we define its semantics properly, it won't be too bad.

EduardoGoulart1:
My suggestion is to focus on this PR (pickling of FAs) and make an issue where we can discuss support for `allow_partial` more deeply. I would like to take a look at the code before making a suggestion. Unfortunately, it's hard to find free time with conference deadlines, etc :confounded:

tests/test_serialization.py
```diff
@@ -0,0 +1,23 @@
...
+import tests.test_fa as test_fa
+
+
+class TestSerialization(test_fa.TestFA):
+    """
+    This tests verifies that the cycle FA -> .pkl -> FA works.
+    The test only applies to DFA and NFAs as other classes do not implement equality
+    """
+
+    def test_serialize_dfa(self):
```
caleb531, line 15: @EduardoGoulart1 Can you please add a docstring here? Something like:

```python
"""Should convert DFA to pickled representation"""
```

tests/test_serialization.py
```diff
@@ -0,0 +1,23 @@
...
+        s = pickle.dumps(self.dfa)
+        dfa = pickle.loads(s)
+        self.assertEqual(self.dfa, dfa)
+
+    def test_serialize_nfa(self):
```
caleb531, line 20: @EduardoGoulart1 Can you please add a docstring here? Something like:

```python
"""Should convert NFA to pickled representation"""
```

automata/base/automaton.py
```diff
@@ -39,6 +39,12 @@ def __delattr__(self, name):
         """Set custom delattr to make class immutable."""
         raise AttributeError(f'This {type(self).__name__} is immutable')
 
+    def __getstate__(self):
```
caleb531, line 42: @EduardoGoulart1 Can you please add a docstring here?

automata/base/automaton.py
```diff
@@ -39,6 +39,12 @@ def __delattr__(self, name):
...
+        return self.input_parameters
+
+    def __setstate__(self, d):
```
caleb531, line 45: @EduardoGoulart1 Can you please add a docstring here?

caleb531 (changes requested):
@EduardoGoulart1 Looks pretty good! Just need a few docstrings for consistency, and we'll be set.

EduardoGoulart1:
> @EduardoGoulart1 Looks pretty good! Just need a few docstrings for consistency, and we'll be set.

Hopefully it's good to go now

caleb531 (approved):
@EduardoGoulart1 Yep, looks good to me. Thank you!

Issue #124, opened 2023-01-12
Visual Automata


lewiuberg:
Hi!

Would you consider implementing the Visual Automata wrapper directly into automata-lib and deprecating Visual Automata?

caleb531:
@lewiuberg So I've been thinking about this ever since you posted the question here, and to be honest, I'm still rather undecided. But there have been a few questions running through my mind:

1. How much do you need to change the visualization code? I am concerned about maintainability if we integrate the visual automata functionality into this codebase. In other words: will this notably increase the maintenance burden, or can we mostly leave the code alone once we integrate it?
2. How much code would need to be integrated? And for which automaton types do you already have this implemented?
3. How much customization do you provide with regard to colors or other styles? Because developers are otherwise effectively committing to whatever styles you've picked for the visuals.

And then the other consideration for me has been whether visualization _should_ really be part of the scope of this library. I know I accepted the `DFA.show_diagram` contribution several years ago, but in retrospect, I've had mixed feelings about its being-part the library overall. I suppose it may be practical because DFAs are common, but I wonder if we open a can of worms and inflate the library substantially by implementation visualization methods for other automaton subtypes (especially if they're only used by a small subset of this library's user base).

@eliotwrobson @abhinavsinha-adrino What do you guys think?

eliotwrobson:
So glancing at the visual automata code, it looks like the main extensions to the core library are expanded versions of the `show_diagram` functions for DFA and NFA. Since the base version of `show_diagram` is already a part of this library, I don't think it's really a huge stretch to include the functionality in visual automata (assuming we can keep the external dependencies somewhat trimmed down). A lot of the other functionality in visual automata should be able to be integrated relatively easily with the code here after the functionality we added in v7. @lewiuberg is my line of thinking correct here?

lewiuberg:
Good morning @caleb531 and @eliotwrobson!

1. I would not have to change the visualization part of the code much I believe. And that is the main motivation for this request. The "wrapper" part of the code however will require much more work to keep up with `automata-lib`; without bringing anything new or improved to the table.
2. I cannot say exactly, but I would say from 500 to 900 lines including code documentation. It all depends, I may be able to make it more efficient since I could do it directly in the source, and I may be able to implement it more efficiently now compared to two years ago. I have implemented `nfa` and `dfa`. I was going to do more, but I could not find the time with work, my wife, and my kids.
[...]

PR #129, opened 2023-03-17
Added jupyter notebook integration and new visualization

khoda81:
## Changes in this PR
- New visualization
  - States are circles instead of ellipses
  - Final state is indicated with a doublecircle
  - Initial state is indicated with a horizontal arrow
  - Multiple transitions between two states are grouped in a comma-separated label
  - Added support for SVG and JPEG
- Jupyter integration
  - Automatically display the SVG representation of FA
  - Support for "image/jpeg", "image/png" and "image/svg+xml" mime-types

## Examples of visual changes
| Old| New |
| :----:|:------:|
| old dfa | new dfa |
| N/A (Resulted in an error, the generated dot was not valid)| dfa-minify

## More Example images
[images.zip](https://github.com/caleb531/automata/files/11006029/images.zip)
The ones not included in the old folder resulted in error when generating.

Note: Tests for the pydot graph need to be updated. I am waiting for these new visualizations to be approved and finalized first.


eliotwrobson:
@khoda81 This is really cool! I haven't looked at the code super closely, but would it be possible to use this Jupyter integration as part of an optional dependency, as in the discussion here: https://github.com/caleb531/automata/issues/124

I think that it would be really nice to have the changes here accompany a move towards the architecture described in the issue above.

khoda81:
@eliotwrobson Looking at [Visual Automata](https://github.com/lewiuberg/visual-automata) seems like I reinvented the wheel. I wish it were mentioned in the project [readme](https://github.com/caleb531/automata/blob/main/README.md).

I agree that `DFA`/`NFA`/`GNFA` classes need some refactoring action; the way I implemented this visualization is by adding [IPython integration](https://ipython.readthedocs.io/en/stable/config/integrating.html) in the `FA` class and implementing a few methods (like `iter_states`, `iter_transitions`, `is_accepted`, ...) in the `DFA`/`NFA`/`GNFA` classes to refactor the visualization logic. The visualization code does not care about what is being visualized it just calls the iter_* methods. This also makes it really easy to add visualization for PDAs and others. This is (for now) lacking from the [Visual Automata](https://github.com/lewiuberg/visual-automata). Refactoring this logic would cut the size of the code almost in half. 

We can add [Visual Automata](https://github.com/lewiuberg/visual-automata) as an optional dependency, then surround the import with a try-except block; just like it is done in many libraries like [flask](https://github.com/pallets/flask/blob/7ed89d3f9d2207c9a607f5dcdce106c0278e1332/src/flask/app.py#L1584). The rest is just replacing my visualizations in the FA class with Visual Automata ones and printing a warning if the import failed and falling back to the default repr of the class. This way the only code added to this library would be only a few lines of code in the FA class, with a few utility methods (`iter_states`, `iter_transitions`, `is_accepted`, ...) in `DFA`/`NFA`/`GNFA`.

This would also remove the need for the `VisualAutomata`/`Visualizer` class for visualizations because the IPython integration methods are only called by Jupyter. So there is no need to wrap the DFA and it will use the visualization automatically when used in Jupyter and act just like it did before everywhere else and when Visual Automata import failed.

If my rough explanations made sense and this is what you expected, I will implement these changes.

eliotwrobson:
@khoda81 I think the way you refactored this is definitely more towards the direction that I think we want to go in with respect to the visualizations described in that issue. Instead of adding Visual Automata as a dependency, this would be a great opportunity to start to integrate that code with your refactors in here (since the purpose of that original issue was to move Visual Automata into this library). As you stated, with your refactors, the size of the code will decrease significantly.

@caleb531 do you have any input?

caleb531:
@eliotwrobson @khoda81 I am hesitant about adding Visual Automata as a dependency because it would duplicate all of the non-visualization Automata code. So I am more in favor of integrating the code into this library.

While I do like my `Visualizer` class proposal from #124, I also want to consider what is most practical for people using this library. If this PR provides the right IPython / Jupyter integrations (i.e. something more practical than the Visualizer class idea), then I'd be fine with that direction. Especially considering that I am not very familiar with IPython or Jupyter.

I suppose if we nix the Visualizer class, I will again raise my concern about the DFA/NFA classes becoming too bloated, but I also want to recognize that [practicality beats purity](https://peps.python.org/pep-0020/).

Also cc'ing @lewiuberg on this thread so that he's in the loop about this proposed new direction, especially since he was already supposedly starting on the Visualizer implementation for #124. @lewiuberg, if you have any thoughts on this, please feel free to chime in.

lewiuberg:
Hi guys! First, sorry for being a bit on/off. I have had a crazy period at work. I have been working on this from time to time, but since automata isn't really something that affects me in my work the direction we decide on isn't that important to me. If you guys want to "steal" the code or add it as a dependency it is fine by me. I just feel it is very beneficial for learning purposes and I don't want all the time I invested into it to go to waste. Although I think it would be the most effective to implement it to automata-lib in terms of maintenance.

One cool thing I want to share is that I just spoke to a professor at Colombia University, and they are now starting to use Visual-Automata in their classes 😃

eliotwrobson:
> One cool thing I want to share is that I just spoke to a professor at Colombia University, and they are now starting to use Visual-Automata in their classes 😃

That's outstanding! That's actually part of why I really like this PR, the integration with something like Jupyter makes it much easier to create self-contained course content using this library.

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+
+        if isinstance(state_data, (set, frozenset, list, tuple)):
+            inner = ", ".join(FA.get_state_name(sub_data) for sub_data in state_data)
+            if isinstance(state_data, (set, frozenset)):
+                return '{' + inner + '}'
+
+            if isinstance(state_data, tuple):
+                return '(' + inner + ')'
+
+            if isinstance(state_data, list):
```
eliotwrobson, line 41: I don't think this case is needed, lists aren't hashable and can't be put in a Python set.

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+            interpolation = Color.interpolate([start_color, end_color], space='lch')
+            # TODO why cycle?
+            color_gen = itertools.cycle(
+                interpolation(x / number_of_colors)
+                for x in range(number_of_colors + 1)
+            )
+
+            # Define all transitions in the finite state machine with traversal.
+            counter = 0
+            for pair in taken_transitions_pairs:
```
eliotwrobson, line 193:
```suggestion
            for from_node, to_node in taken_transitions_pairs:
```

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+                counter += 1
```
eliotwrobson, line 194: Instead of this, you can use an `itertools.count` and call `next` instead of having to increment manully.

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+                edge_color = next(color_gen)
+                graph.edge(
+                    pair[0],
+                    pair[1],
+                    label=" [{}]\n{} ".format(counter, pair[2]),
```
eliotwrobson, line 199:
```suggestion
                    label=f" [{counter}]\n{pair[2]} ",
```
a bunch of these `.format` calls can be replaced with f-strings.

automata/fa/nfa.py
```diff
@@ -238,6 +237,7 @@ def eliminate_lambda(self):
 
     def _check_for_input_rejection(self, current_states):
         """Raise an error if the given config indicates rejected input."""
+        # TODO isn't set.isdisjoint() faster?
         if not (current_states & self.final_states):
```
eliotwrobson, line 241: Not sure if it's faster but definitely cleaner + more expressive, let's do it!
```suggestion
        if current_states.isdisjoint(self.final_states):
```

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+                label = ",".join(sorted(labels))
+                graph.edge(
+                    from_node,
+                    to_node,
+                    label=label,
+                    arrowsize=arrow_size,
+                    fontsize=font_size,
+                )
+        else:
+            # TODO
```
eliotwrobson, line 165: Maybe the way to do this is to just call `read_input_stepwise` from the parent class? I know this won't work for the GNFA but I don't think it makes sense to support an input string there. It would be cool to visualize the steps in the GNFA -> regex algorithm though.

eliotwrobson (commented):
Some comments, looks really great so far!

automata/fa/nfa.py
caleb531: @eliotwrobson @khoda81 Just merged #130, so you can remove the TODO here. When that's done, please feel free to mark this comment thread as _Resolved_.

automata/fa/fa.py
caleb531: @khoda81 Agreed with @eliotwrobson. Also, please prefer single quotes instead of double quotes where possible for strings, since the former is what's used in most of the codebase.

automata/fa/fa.py
caleb531: Agreed with @eliotwrobson on this.

automata/fa/fa.py
caleb531: Agreed with @eliotwrobson on this, because tuple unpacking is awesome.

automata/fa/fa.py
caleb531: @eliotwrobson While I agree that `itertools.count` would be a more generalized/idiomatic alternative, we already plenty of other places where manually-incremented counters are used. Therefore, I'm not as much of a stickler about which is used at the moment.

@eliotwrobson If I may, I'm going to resolve this particular comment thread so @khoda81 can focus on the other items.

automata/fa/fa.py
caleb531: @eliotwrobson Well, `GNFA.read_input_stepwise` already raises the appropriate `NotImplementedError`, so I think it would be fine to call `read_input_stepwise` here. cc @khoda81

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+
+    @abc.abstractmethod
+    def is_accepting(self, state):
+        """Check if a state is an accepting state."""
+
+    @abc.abstractmethod
+    def is_initial(self, state):
+        """Check if a state is an initial state."""
+
+    def show(
```
caleb531, line 65: @khoda81 Why is this method called `show` and not `show_diagram`? I would much prefer the latter, and not sure why `show_diagram` was removed. cc @eliotwrobson

caleb531:
@khoda81 This is a great start so far! I do have a few points of constructive feedback:

1. I've tried to chime in on all the comments that @eliotwrobson left for you. I agree with most of his feedback, and would ask that you implement those suggestions of his on which I have indicated approval.
2. The automated tests are currently failing. Please make sure to run `nose2` while you are developing and fix any broken tests (or, as needed, add new tests to maintain code coverage)
3. I think my recent merge of #130 has caused a conflict in this branch, so that will need to be resolved as well before I am able to merge this PR

automata/fa/fa.py
```diff
@@ -2,11 +2,245 @@
...
+        self,
+        input_str: str | None = None,
+        save_path: str | os.PathLike | None = None,
```
eliotwrobson, line 68: @caleb531 somewhat tangentially related, but how would you feel about trying to add type annotations to the library again? I know this got dropped a while ago, but I'm realizing it makes writing code easier, since editors like VSCode use these annotations for autocomplete features. The annotations don't have to be super detailed, just simple things like "-> DFA" or "-> Set" and things like that.

caleb531: @eliotwrobson You know, I was actually thinking about this the other day when I noticed type annotations in this PR lol.

I recall one of the places we left off (in #43) was [an issue you were having](https://github.com/caleb531/automata/pull/43#issuecomment-1101880464) around changing function signatures in derived classes, and mypy not being happy with that. But perhaps the tooling is better now (especially with a different type checker like Pylance) to where that won't be much of an issue.

But yes, I like the idea of trying to integrate types with this library, especially since it _is_ a library and not just application code. And if we have a reasonable solution to the above issue, then I think we'll be golden.

Is this something you would be up for taking on again?

eliotwrobson: > I recall one of the places we left off (in #43) was [an issue you were having](https://github.com/caleb531/automata/pull/43#issuecomment-1101880464) around changing function signatures in derived classes, and mypy not being happy with that. But perhaps the tooling is better now (especially with a different type checker like Pylance) to where that won't be much of an issue.

I think that Pyright (the typechecker used by Pylance) will also complain about this (it violates Liskov), but this might be as simple as using a wrapper function. The main issue was adding annotations that had issues with 3.6, but that's no longer an issue.

> Is this something you would be up for taking on again?

Sure! I can take a look and start playing around.

automata/fa/fa.py
khoda81: Technically, the `set` case a few lines above is also unnecessary, but it's better to support them in case someday the hashable-ity requirement of fa-states changes someday or when this visualization is used for other automata. I will do a performance benchmark on all of the code before merging. If there are any notable hits, I will remove this.

automata/fa/fa.py
khoda81: True, I didn't want to touch this code until I find a way to make the step visualizations work.

automata/fa/nfa.py
khoda81: It might be faster because when taking `a & b` you are basically doing this: 
```python
{x for x in a if x in b}
```
Which calculates the full intersection to see if there is any intersection. But we can terminate early when we find the first element that is contained in both, so `set.isdisjoint` is kinda like this:
```python
def isdisjoint(a, b):
    for x in a:
        if x in b:
            return False
    return True
```
In theory, it should be faster on large sets, but things might be different in the real-world...

automata/fa/fa.py
khoda81: That was a temporary change to slowly move the code from `show_diagram` to this new method. I will change the name in the back to `show_diagram` in the next commit.

automata/fa/dfa.py
```diff
@@ -1279,43 +1278,19 @@ def get_name_original(states):
...
-        return graph
+    def iter_transitions(self):
+        return (
+            (from_, to_, symbol)
+
+            for from_, lookup in self.transitions.items()
+            for symbol, to_ in lookup.items()
+        )
+
+    def is_accepting(self, state):
```
caleb531, line 1292: @khoda81 Do we need this method? I feel like `state in self.final_states` would be clearer while not requiring that many more characters.

automata/fa/dfa.py
```diff
@@ -1279,43 +1278,19 @@ def get_name_original(states):
...
+        return state in self.final_states
+
+    def is_initial(self, state):
```
caleb531, line 1295: @khoda81 Same question as above about this method. I personally think it would be clearer to inline.

automata/fa/dfa.py
khoda81: Yes, and potentially other automatas.

automata/fa/dfa.py
khoda81: I wanted to make the `show_diargram` method not interact and make assumptions about what it is visualizing, and be able to work with just these 4 methods. This just makes it easier later on when adding visualization for other automata.

caleb531:
@khoda81 Just a heads up—for more consistent code style across the entire codebase, I plan to integrate the [black](https://github.com/psf/black) auto-formatter and apply it to all Python files in the library.

Once I do this on the `develop` branch, I will update this PR to auto-format the files you've already modified. This should prevent future formatting conflicts, and also eliminate some feedback loops in this PR for pedantic topics like quote style. 😅

cc @eliotwrobson

caleb531:
@khoda81 Okay, I've added black to the `develop` branch, and have rebased your feature branch (`jupyter-integration`) off of my official `develop` branch. Please make sure to pull in these new (rebased) changes properly.

I've done my best to resolve conflicts and preserve the integrity of your changes, but please double-check just to be sure.

cc @eliotwrobson

khoda81:
@caleb531 black is a great addition! It would be cool to add it as a pre-commit hook. I might do that after this PR. It would help greatly with contributions. I usually use [`Gymnasium`'s](https://github.com/Farama-Foundation/Gymnasium/blob/main/.pre-commit-config.yaml) pre-commit hooks as a template for my own projects.

caleb531:
> It would be cool to add it as a pre-commit hook. I might do that after this PR. It would help greatly with contributions. I usually use [Gymnasium's](https://github.com/Farama-Foundation/Gymnasium/blob/main/.pre-commit-config.yaml) pre-commit hooks as a template for my own projects.

@khoda81 So I have limited knowledge of pre-commit hooks, but from what I understand, they can't be committed to the repository directly and require some manual setup to activate. So it seems like the pre-commit hook is only good for those contributors who take the time to set it up. I guess the same could be said of editor plugins.

In your experience, is there a good way to streamline the management/adoption of pre-commit hooks for a project?

cc @eliotwrobson (in case you have any thoughts on this)

khoda81:
@caleb531 To add them to this project we need to only add the `.pre-commit-config.yaml` file to the root of the project, which contains the hooks (like black, trailing-whitespace, isort, etc) and configurations related to them.
What changes for contributors is explained in the contribution file [here](https://github.com/Farama-Foundation/Gymnasium/blob/main/CONTRIBUTING.md#git-hooks) (which this project also needs, btw). The setup is pretty minimal, pre-commit can be pip installed once and most hooks would also be automatically set up once. Visit [pre-commit.com](https://pre-commit.com/) for more info.

caleb531:
@khoda81 This project _does_ have a CONTRIBUTING file, but it's under the `.github` subdirectory:

https://github.com/caleb531/automata/blob/develop/.github/CONTRIBUTING.md

Anyway, you've persuaded me that a pre-commit hook would be a valuable addition to this project. I would greatly appreciate a separate PR for this if you have the time to contribute one.

khoda81:
@caleb531 my bad, I somehow missed the file. I will add pre-commit hooks in a separate PR.

khoda81:
@eliotwrobson @caleb531
I have been playing around with different implementations of the step visualization feature and I have two ways of achieving it:
1. Adding an alternate method to DFA/NFA that returns the steps with a data structure that is easier to work with for visualization
2. Adding a switch case for DFA/NFA in the `show_diragram` method to translate the result of `read_input_stepwise` based on the class that we are working with.

The first solution is more on brand with the rest of this PR but would add another method for the purpose of visualization which I have been trying to avoid. The upside would be that if implemented properly it can replace `read_input_stepwise` and gives a more generalized API for FAs and makes it easier to add visualizations for PDAs or any custom automata as well. The second solution though would be probably less code and using the existing code.
We can also use the second option as a temporary solution and play around with different generalizations of FAs in a separate PR.
Any thoughts on this?

eliotwrobson:
@khoda81 For option 1, do you mean that the method would return a data structure holding information about each step that is taken, and then the visualization code could just read this directly? Does this differ greatly from just yielding all results from the generator given by `read_input_stepwise`? I don't have an immediate strong opinion either way.

caleb531:
@khoda81 I think any major re-architecture of the input-reading APIs will require some iteration and proof-of-concepts, for sure. On that note, I like the idea of starting with the second option since it seems like it would have less of an API impact, and would be a good starting point to explore larger changes.

Out of curiosity, what kinds of detail is missing from the current input-reading API that a new API would be more useful in providing?

khoda81:
@caleb531 @eliotwrobson My proposal for the new API is here #138.
I need the read_input_stepwise to tell me which edges were used in every single transition. I will handle `NFA`/`DFA` input read separately using a good old if-else for now...

BTW, I have a hard time figuring out what the purpose of `GNFA` class is. Does not seem to be used anywhere internally.

khoda81:
Step visualizations are implemented and seem to be working properly. I am torn between different ways to label transitions in a way not to make the diagram too busy and concise at the same time.
step_viz1
step_viz3
step_viz5
@eliotwrobson @caleb531 any opinions/proposals?

eliotwrobson:
@khoda81 I think it's always going to be a bit tricky visualizing D/NFAs reading input much longer than their number of states (inherently things are going to get cluttered, it would be better to have an animation rather than a static image but that's way out of scope I think). I like the way you have the color grading become more green as more of the input is being read. Maybe try italicizing the `[#1]` labels so they stand out from just the input symbols? Overall things look good though.

Could you try an example with a somewhat larger DFA constructed using the `from_finite_language` method? I think it might be a common use case for people.

I haven't looked at the code, but it seems like a few of the backend changes might overlap a bit with https://github.com/caleb531/automata/pull/139, and this branch is already bit out of date with `develop`. Once that PR is merged, the next step is to rebase the changes here and see if any logic can be condensed (cc @caleb531 to see if we can get that PR merged soon, since it's already been approved).

khoda81:
@eliotwrobson 
> I like the way you have the color grading become more green as more of the input is being read

Credit for that Idea goes to lewiuberg. It also goes from yellow to red if the input is not accepted.
| | |
|:-:|:-:|
| Italic | language_accepted_b | 
| Bold | language_accepted_i | 
| Italic & Bold | language_accepted_b_inner_i_new |

### And here is a rejection example
language_failed_b_inner_i_new

### Here is my playground notebook + diagrams
[Playground.zip](https://github.com/caleb531/automata/files/11190162/Playground.zip)

eliotwrobson:
@khoda81 #139 was recently merged, you should be good to rebase!

khoda81:
@eliotwrobson With `NFA` input visualization implemented the only remaining thing to add is the table feature, which includes:
1. `fa.table`: returns a table of all transitions
2. `fa.input_check(input)`: returns the history of states when reading input

You can see examples of this on [VisualAutomata's readme](https://github.com/lewiuberg/visual-automata).
The only problem is that it uses pandas which means to add this feature we have to add pandas as a dependency. So is it worth it?

caleb531:
@khoda81 @eliotwrobson From the sound of it, and after reviewing the VisualAutomata docs, I am not a huge fan of the table features (both `fa.table` and `fa.input_check`), since we are hardcoding the implementation to a specific style/format, and unless we provide a lot of options to control the display, the end user is probably going to write their own implementation anyway.

But if you want to discuss it further, I would prefer a separate PR anyway.

khoda81:
@eliotwrobson The diff problem should be fixed now, I forgot to sync my fork before rebasing.
I am starting to fix the tests and add visualization documentation.

eliotwrobson (changes requested):
Starting to look good! Overall, I think we're getting ahead of ourselves with some of the abstractions introduced, and I think some functions can be simplified.

automata/base/automaton.py
```diff
@@ -124,29 +123,10 @@ def copy(self) -> Self:
         """Create a deep copy of the automaton."""
         return self.__class__(**self.input_parameters)
 
-    # Format the given value for string output via repr() or str(); this exists
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson @khoda81 I originally wrote this function because calling `repr()` on an automaton instance was producing things like `frozenset({ ... })` and `frozendict({ ... })` in the output, but I wanted to display something friendlier like `{ ... }` and `{ "key": ... }` in the repr output. I thought this would be especially acceptable considering that native sets and dicts get type-cast to their frozen variants on instantiation anyway, so the repr output is still consistent with the code to create that instance.

I honestly don't know why this was removed, but it's my fault because I didn't say anything lol. I think it is probably worth keeping, though.

automata/fa/fa.py
```diff
@@ -12,3 +20,217 @@ class FA(Automaton, metaclass=abc.ABCMeta):
...
+        Iterate over all transitions in the automaton. Each transition is a tuple
+        of the form (from_state, to_state, symbol)
+        """
+
+    @abc.abstractmethod
+    def is_accepting(self, state: FAStateT) -> bool:
+        """Check if a state is an accepting state."""
+
+    @abc.abstractmethod
+    def is_initial(self, state: FAStateT) -> bool:
```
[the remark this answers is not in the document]
caleb531: @khoda81 I agree with @eliotwrobson. I would prefer to drop this method for this PR.

automata/fa/fa.py
```diff
@@ -12,3 +20,217 @@ class FA(Automaton, metaclass=abc.ABCMeta):
...
+            if isinstance(state_data, list):
+                return "[" + inner + "]"
+
+        return str(state_data)
+
+    def get_edge_name(self, symbol) -> str:
+        return str(symbol)
+
+    @abc.abstractmethod
+    def iter_states(self) -> Iterable[FAStateT]:
```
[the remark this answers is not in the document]
caleb531: @khoda81 Also agreed with @eliotwrobson about this. I would prefer to just write `for state in states:` rather than having an intermediary method.

automata/fa/gnfa.py
```diff
@@ -320,39 +319,16 @@ def option(self):
...
-        return graph
+    def iter_transitions(self):
+        return (
+            (from_, to_, symbol)
+            for from_, lookup in self.transitions.items()
+            for to_, symbol in lookup.items()
+            if symbol is not None
+        )
+
+    def is_accepting(self, state):
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson You know, I've been thinking about this ever since [I first asked](https://github.com/caleb531/automata/pull/129/files#r1153860767) about the need for `is_accepting` and realized that it was for things like GNFA.

Firstly, I would say that the [formal definition of a GNFA](https://en.wikipedia.org/wiki/Generalized_nondeterministic_finite_automaton#Formal_definition) stipulates that a GNFA has one and only one accept state, which I would prefer not to stray from. However, I think I have a solution that will work for everyone:

My proposal is to add a `@property`-type member to GNFA called `final_states` , which simply returns a `frozenset` containing that single final state. Something like:

```python
@property
def final_states(self):
    return frozenset({self.final_state})
```

The nice thing about this is because `frozenset` is hashable, `final_states` will always return something that is comparably the same on each access (hence why I didn't bother to suggest we use `@cached_property` or something).

automata/fa/fa.py
```diff
@@ -12,3 +20,217 @@ class FA(Automaton, metaclass=abc.ABCMeta):
...
+        """Iterate over all states in the automaton."""
+
+    @abc.abstractmethod
+    def iter_transitions(self) -> Iterable[tuple[FAStateT, FAStateT, Any]]:
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson ...and I think I can see why, namely because the transition maps for the various automaton types are all slightly different. DFAs don't use sets, but NFAs do, GNFAs are just plain weird lol, and I haven't even gotten to PDAs and TMs yet. lol

So yep, I agree that `iter_transitions` is fine to keep! Resolving this thread.

automata/fa/fa.py
```diff
@@ -12,3 +20,217 @@ class FA(Automaton, metaclass=abc.ABCMeta):
...
+                directory=directory,
+                filename=filename,
+                format=format,
+                cleanup=cleanup,
+                view=view,
+            )
+
+        return graph
+
+    def _get_input_path(self, input_str):
```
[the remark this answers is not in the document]
caleb531: @khoda81 @eliotwrobson Agreed that this method should be abstract.

pyproject.toml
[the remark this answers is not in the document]
caleb531: @eliotwrobson @khoda81 Yes, it seems there is a way to specify optional dependencies!

```toml
# pyproject.toml
# ...
[project.optional-dependencies]
PDF = ["ReportLab>=1.2", "RXP"]
```

Source: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

pyproject.toml
```diff
@@ -13,7 +13,9 @@ maintainers = [
     {name = 'Caleb Evans', email = 'caleb@calebevans.me'}
 ]
 dependencies = [
-    "pydot>=1.4.2",
+    "coloraide>=1.8.2",
+    "frozendict>=2.3.4",
+    "graphviz>=0.20.1",
```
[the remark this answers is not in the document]
caleb531: @eliotwrobson @khoda81 Agreed—please see my comment about for how to specify optional dependencies (at least according to my quick Google search 😅).

automata/fa/fa.py
```diff
@@ -12,3 +20,217 @@ class FA(Automaton, metaclass=abc.ABCMeta):
...
+        """Calculate the path taken by input."""
+
+        raise NotImplementedError(
+            f"_get_input_path is not implemented for {self.__class__}"
+        )
+
+    def _ipython_display_(self):
```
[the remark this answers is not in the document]
khoda81: The return value of this function is ignored by ipython.

automata/base/automaton.py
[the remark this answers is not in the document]
khoda81: @eliotwrobson @caleb531 Can we remove this function in the favour of `get_state_name(state)`? or remove `get_state_name(state)` in the favour of this? I think there is some duplicate logic here.

automata/fa/nfa.py
```diff
@@ -1024,3 +985,65 @@ def add_any_transition(start_state_dict, end_state):
...
+
+    def get_edge_name(self, symbol):
+        return "ε" if symbol == "" else str(symbol)
+
+    def _get_input_path(self, input_str):
+        """Calculate the path taken by input."""
+
+        visiting = set()
+
+        def gen_paths_for(state, step):
```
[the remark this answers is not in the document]
khoda81: @eliotwrobson I might have overthought this one! The reason for the two mutually recursive functions is to find a preferably accepted path with the fewest jumps and longest input read if the input was not accepted. This can be simplified with some hit to the path found (trying to find the first path for example). The `read_input_stepwise` method does not share a lot of logic with these two functions. We don't need to keep track of the full state of nfa when performing a pathfinding algorithm (DFS). It would have been better if `read_input_stepwise` were to take `starting_state` as an argument instead of using `initial_state` as the starting state all the time. And even then we still would have needed all the transitions used in each jump which we don't get.

automata/fa/fa.py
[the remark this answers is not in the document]
khoda81: @eliotwrobson is it okay to turn them into abstract properties instead?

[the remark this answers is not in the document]
khoda81: @eliotwrobson Is it okay to turn them into abstract properties instead?

automata/base/automaton.py
```diff
@@ -22,10 +21,10 @@ class Automaton(metaclass=abc.ABCMeta):
     __slots__: Tuple[str, ...] = tuple()
 
     initial_state: AutomatonStateT
-    states: AbstractSet[AutomatonStateT]
```
khoda81, line 25: @eliotwrobson Is it okay to use Set instead of AbstractSet? They are equivalent and AbstractSet is defined as an alias for Set.

eliotwrobson: I don't think this is true (see https://stackoverflow.com/a/50167430/2923069). The point if using `AbstractSet` is that both `Set` and `FrozenSet` can be used for this.

automata/base/automaton.py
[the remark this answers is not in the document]
eliotwrobson: If you can find a way to condense some of the logic that's fine, but the `get_state_name` function has a fundamentally different purpose than `_get_repr_friendly_value`, so I'm ok with having both if they're both short.

automata/fa/nfa.py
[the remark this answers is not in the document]
eliotwrobson: @khoda81 that makes sense. Not a huge deal if these two functions can't share any logic (as you said, `read_input_stepwise` only really cares about the current set of states not the exact transitions taken), but it would be nice to implement this function iteratively. I think finding the simplest path should be doable with BFS?

Also, be sure to test these functions with an NFA that has a cycle of lambda transitions, I think it might behave strangely in that situation.

automata/fa/fa.py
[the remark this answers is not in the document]
eliotwrobson: The use of `abstractproperty` is actually deprecated (see https://docs.python.org/3/library/abc.html#abc.abstractproperty). I think it's unnecessary since these are always set in the `__init__` function at no cost, so there's not really a benefit to making them properties.

automata/fa/dfa.py
```diff
@@ -1553,38 +1550,39 @@ def get_name_original(states: FrozenSet[DFAStateT]) -> DFAStateT:
             final_states=dfa_final_states,
         )
 
-    def show_diagram(self, path=None):
+    def iter_transitions(self):
```
eliotwrobson, line 1553: This function should have a type annotation.

automata/fa/dfa.py
```diff
@@ -1553,38 +1550,39 @@ def get_name_original(states: FrozenSet[DFAStateT]) -> DFAStateT:
...
+            input_str (str): The input string to run on the DFA.
+
+        Returns:
+            tuple[list[tuple[DFAStateT, DFAStateT, DFASymbolT], bool]]: A list
+            of all transitions taken in each step and a boolean indicating
+            whether the DFA accepted the input.
+
         """
-        # Nodes are set of states
+        state_history = [
```
eliotwrobson, line 1575:
```suggestion
        state_history = list(self.read_input_stepwise(input_str, ignore_rejection=True))
```
You do a similar comprehension in a few places, I think this can be used to simplify things.

automata/fa/nfa.py
```diff
@@ -1024,3 +984,82 @@ def add_any_transition(start_state_dict, end_state):
...
+
+    def iter_transitions(self):
+        return (
+            (from_, to_, symbol)
+            for from_, lookup in self.transitions.items()
+            for symbol, to_lookup in lookup.items()
+            for to_ in to_lookup
+        )
+
+    def get_edge_name(self, symbol):
```
eliotwrobson, line 996: I think this should be made an abstract method in the base class?

automata/base/automaton.py
caleb531: @khoda81 I would prefer `AbstractSet`, since that is the convention @eliotwrobson and I have settled on for this project (given the reasons that Eliot linked to in his comment).

automata/base/automaton.py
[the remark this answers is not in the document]
caleb531: @khoda81 So reviewing your code, it looks like `get_state_name` deals with a single state, but the purpose of `_get_repr_friendly_value` is more oriented towards handling `set` and `dict` structures.

But besides that, `get_state_name` does not look like it's designed for repr output, but rather, an aesthetically-pleasing display of the state used in an image of an automaton.

For example, one of the return values for `get_state_name` is `λ`, which makes sense when you are producing a rendered image of the machine. However, because `λ` is not considered the empty string when constructing an Automaton in Python code, `λ` would not make sense to use for repr output (because according to the Python docs, repr output should match the Python code to re-construct that object).

All that is to say, because the `get_state_name` is not designed for repr output and only operates on a single state, it has a fundamentally different purpose than `_get_repr_friendly_value`.

cc @eliotwrobson

eliotwrobson:
@khoda81 I just pushed some minor changes to the develop branch, so there are a couple of small merge conflicts that have to be resolved. It should be pretty straightforward to address those along with the remaining comments on here (I don't think there are _too_ many).

Feel free to let me know if you are busy and would like some help. It would be nice to have this merged soon, since I think we're getting close to moving through enough items to get to a v8 release.

khoda81:
@eliotwrobson Sorry for the delay, I am taking a break for now, I'm going through university midterms rn being an undergrad. :)
I will push some minor changes that were remaining to this branch for now but absolutely go ahead if you can. It might take another week or two for me to come back. All planned features are implemented, this pr just needs some finishing touches...

eliotwrobson:
@khoda81 no worries! I'm a bit busy this week too wrapping up my semester. I'll try to get some changes on this branch after that, since I'm shooting to get most of the work done for the next release before I go on vacation.

eliotwrobson:
@khoda81 I made some pushes, so be sure to pull. The main changes were addressing the items listed on the discussions and cleaning up some types. I think the main changes remaining are to polish up the handling of optional dependencies and update the test cases to increase the code coverage.

caleb531:
@eliotwrobson 

>  It only requires one extra line to get that to work, and not having that removes the dependence on IPython entirely (which may be desirable). Curious what others think (cc @caleb531)

I like the idea of keeping it simple and avoiding wrangling in large dependencies when they could be easily added on a per-user basis.

I also agree about the switch to `pygraphviz` for the reasons you presented.

@eliotwrobson Is there anything more you plan to add to this PR? And @khoda81 was there anything more you want to review here or discuss based on Eliot's recent changes?

eliotwrobson:
@caleb531 

> Is there anything more you plan to add to this PR? And @khoda81 was there anything more you want to review here or discuss based on Eliot's recent changes?

So in terms of features, I think that what's here is fine. The issue is that to remove the dependency and switch to `pygraphviz` over `graphviz`, I had to remove the feature where the diagram was displayed directly in the notebook on the `show_diagram` call. This is because the `pygraphviz` graph doesn't have Jupyter notebook integration similar to that of `graphviz`.

To get around this, I looked at a previous commit where @khoda81 created a `Ipythongraph` class, that was essentially a wrapper class facilitating a nicer Jupyter display. I think we could add this class back, and have it inherit from the `AGraph` class from `pygraphviz`. Then all that remains is to update the tests and clean things up. I thought about changing the recursive NFA path algorithm to an iterative one, but that can be done later.

I'm a bit busy right now and will be for at least the next week, so if someone wants to pick this up, the changes should be pretty straightforward.

khoda81:
@eliotwrobson I have pulled the changes and my Windows setup is broken (I have not installed pygraphviz yet). The typing at 

https://github.com/caleb531/automata/blob/5c44c0e459dd27035728a6b46a7bafccfdbf7631/automata/fa/fa.py#L79 seems to throw an error when pygraphviz is not installed.
Also, I'm having some issues installing pygraphviz on Windows. A simple pip install does not install pygraphviz but there is a conda channel to install it on Windows, which I prefer not to because it will make the installation process more difficult. Since it is required to be installed separately and adding to dependencies would not help.

eliotwrobson:
@khoda81 if you set `from __future__ import annotations ` that should fix it. But you'll need pygraphviz to be able to work on the tests anyway.

eliotwrobson:
@caleb531 @khoda81 So I've updated the test cases and looks like everything is passing with coverage. I think this should be ready to start reviewing. A few notes:

- I set up optional dependencies, but not 1000% sure they're working correctly. I also didn't update the documentation until after we get this configuration fully settled.
- I removed the `show_None` attribute from the display for `GNFA`, since it seemed kinda pointless to display a bunch of emptyset symbols everywhere.
- I don't know why coveralls is displaying 100% coverage, since there's a bit of new code that's definitely not getting hit. Adding the tests to cover that should be easy but curious why this is being output right now.

pyproject.toml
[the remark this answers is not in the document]
caleb531: @eliotwrobson Looking over this again, were you asking about _importing_ an optional dependency in this file (i.e. in a way that doesn't break the module if it's not installed)?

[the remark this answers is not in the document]
eliotwrobson: I followed directions on the link in the latest commit. I'm not sure how to test whether it works or not, however.

caleb531:
@eliotwrobson @khoda81 Sorry for the delay—I've had a busy week so far. A few points:

1. @eliotwrobson I left you a question about the optional dependencies—just want to make sure we have that cleared up before I merge anything
2. The removal of `GNFA. show_None` is fine with me, and the rationale makes sense
3. I've noticed Coveralls sometimes displays higher coverage than I was expecting, so once this PR is merged, I will run a coverage report myself (locally) to see how it compares

Now, there's a ton of code in this PR, so I don't have time to review _all_ of it today. But on the surface (for what I've skimmed), it looks pretty good.

khoda81:
@caleb531 There is a slight problem with replacing graphviz with pygraphviz. I have a working installation of graphviz binary and graphviz python package. A simple pip install will not work on Windows because of build issues, therefore it needs to be installed with:
```
python -m pip install --global-option=build_ext `
              --global-option="-IC:\Program Files\Graphviz\include" `
              --global-option="-LC:\Program Files\Graphviz\lib" `
              pygraphviz
```
Which as a result becomes part of the installation process of automata lib (on Windows). This is while the graphviz binary needs to be installed separately anyway. And I need to add, automata lib cannot be installed on Windows if we add pygraphviz as a dependency since the installation of pygraphviz would fail. Therefore pip will terminate the entire installation.

Since graphviz did not provide proper API for testing purposes, one solution would be to switch back to pydot. But that might need to remain for another PR.

@eliotwrobson I have tested optional dependencies and it did work for me. Though as mentioned above I needed to install pygraphviz before I did `pip install -e .[visual]`.

eliotwrobson:
@khoda81 I don't think the issue you're getting at isn't as big of a deal in light of the introduction of optional dependencies. Python packages that interface with other applications are always going to be harder to install across different operating systems (this is the reason why, on my windows machine, I used conda to set up pygraphviz). As long as the package works without having to install the optional dependencies, I think we're in good shape.

@caleb531 No worries on the review! I think I might see if I can run coveralls locally and see what happens to try to boost coverage while you're taking a look. WRT optional dependencies, I think that everything works (besides the display function) if visual dependencies are not installed because of the import guard I set up. I wonder if there's a way to set up automated testing for that.

caleb531:
@eliotwrobson 

> I'm not sure how to run coveralls locally...

Oh hmm, I can tell you how to run `coverage` locally, but I've never tried to run Coveralls locally. It's strange though, because the CI job is just generating an `lcov` file, which is then being fed to Coveralls. So  I wonder if Coveralls is actually doing anything beyond parsing the coverage data you feed it.

> WRT optional dependencies, I think that everything works (besides the display function) if visual dependencies are not installed because of the import guard I set up. I wonder if there's a way to set up automated testing for that.

I think this could be accomplished with a second CI job that executes only visualization-related tests with the additional dependencies.

eliotwrobson:
@caleb531 ahh I see, I think coveralls is parsing the data correctly but something about the coverage configuration might be off. I wasn't able to get nose2 coverage running locally.

automata/fa/fa.py
```diff
@@ -1,19 +1,230 @@
...
 
+    @staticmethod
+    def get_state_name(state_data: FAStateT) -> str:
+        """
+        Get an string representation of a state. This is used for displaying and
+        uses `str` for any unsupported python data types.
+        """
+        if isinstance(state_data, str):
+            if state_data == "":
+                return "λ"
```
caleb531, line 41: @eliotwrobson @khoda81 What is the difference between `𝛌` or `ε` in an Automaton Theory context? I've seen both used, sometimes seemingly interchangeably, to represent empty string.

automata/fa/fa.py
```diff
@@ -1,19 +1,230 @@
...
+            - fig_size (tuple, optional): Figure size. Defaults to None.
+            - font_size (float, optional): Font size. Defaults to 14.0.
+            - arrow_size (float, optional): Arrow head size. Defaults to 0.85.
+            - state_separation (float, optional): Node distance. Defaults to 0.5.
+        Returns:
+            AGraph corresponding to the given automaton.
+        """
+
+        if not _visual_imports:
+            raise ImportError("Missing visual imports.")
```
caleb531, line 101: @eliotwrobson @khoda81 I would like this error message to be more descriptive about what specific package must be installed to resolve the error. Maybe something like

```python
if not _visual_imports:
    raise ImportError("Missing visualization packages; please install coloraide and pygraphviz.")
```

pyproject.toml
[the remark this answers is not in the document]
caleb531: @eliotwrobson Gotcha. I think it's safe to close this thread, then. But let me know if there's anything else you want to discuss on this matter.

pyproject.toml
```diff
@@ -13,11 +13,17 @@ maintainers = [
...
-    "frozendict>=2.3.4", #TODO I think that typing extensions needs to be in here
+    "frozendict>=2.3.4",
+    "typing-extensions>=4.5.0"
 ]
 
+# Per https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies
+[project.optional-dependencies]
+visual = ["coloraide>=1.8.2", "pygraphviz>=1.10"]
+
+#TODO maybe add stuff from here?: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/#adding-pytyped
```
caleb531, line 25: @eliotwrobson @khoda81 Just pointing out that per this TODO, we will still want to add `py.typed` per the linked article so that the user's mypy will be able to use our type annotations.

caleb531 (changes requested):
@eliotwrobson @khoda81 Okay, I've reviewed the full PR, and it looks mostly fine. I've left a few comments on leftover TODOs and some small changes. 

The only thing I'm dealing with now is on my Mac (macOS Ventura), I'm having some trouble installing pygraphviz. I _did_ find some code that installs it:

```sh
pip install --global-option=build_ext --global-option="-I$(brew --prefix graphviz)/include/" --global-option="-L$(brew --prefix graphviz)/lib/" pygraphviz
```

However the `--global-option` CLI option is deprecated per pypa/pip#11859, and it isn't clear how the `--config-settings` option is to be used.

automata/fa/fa.py
eliotwrobson: There is no difference. It just depends on which symbol you want to use to represent the empty; I prefer `ε`, though we should probably stick to one convention in this library.

pyproject.toml
eliotwrobson: Is `py.typed` included by default or do we have to set something in here? The file is already in the base directory on the `develop` branch.

automata/fa/fa.py
caleb531: @eliotwrobson That's what I thought, thanks. The class I took in university used "lambda" almost ecxlusively—I didn't learn about "epsilon" as an alternative term until I googled more about the automata theory (after I graduated).

A quick search through the codebase's Markdown files reveals that we seem to lack a lean towards one or the other term (i.e. "lambda" vs. "epsilon").

Screenshot 2023-05-18 at 9 04 09 PM

Screenshot 2023-05-18 at 9 04 01 PM

I suppose you could argue that the name of our existing method `eliminate_lambda` implies that 'lambda' is the convention, but I don't think that's enough to establish a preference. I think that's why I refer to both terms in the documentation often, so that people who are familiar with either convention

I _was_ debating if this would be worthy of being a function parameter or even a global config option, so that users could choose one or the other.

pyproject.toml
caleb531: @eliotwrobson Hmm, a quick Google search on my end makes me think yes, it needs to be included in pyproject.toml as well ([source](https://stackoverflow.com/a/76214408/560642)):

```toml
[tool.setuptools.package-data]
"pkgname" = ["py.typed"]
```

And [here's an example](https://github.com/lsst/utils/blob/main/pyproject.toml#L50) of a project using the above in the wild.

automata/fa/fa.py
```diff
@@ -1,19 +1,230 @@
...
+
+        # Defining the graph.
+        graph = pgv.AGraph(strict=False, directed=True)
+
+        # TODO test fig_size
```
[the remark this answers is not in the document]
eliotwrobson: Yeah, this kinda goes back to the weird behavior of the line coverage; there's definitely stuff that's not getting hit.

automata/fa/fa.py
```diff
@@ -1,19 +1,230 @@
...
+        graph.layout(prog=layout_method)
+
+        # Write diagram to file. PNG, SVG, etc.
+        # TODO gotta fix this
+        if path is not None:
+            save_path_final: pathlib.Path = pathlib.Path(path)
+
+            # directory = save_path_final.parent
+            # directory.mkdir(parents=True, exist_ok=True)
+            # filename = save_path_final.stem
```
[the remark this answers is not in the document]
eliotwrobson: I did not, I wasn't sure if it was still needed once I switched the libraries and didn't get a chance to test it out yet.

pyproject.toml
eliotwrobson: @caleb531 Added this here https://github.com/caleb531/automata/pull/129/commits/b0ce2adaf2f11bc39ea3568e683853e2240f3a03

In that same commit, I deleted the `.coveragerc` file and moved the configuration to the `pyproject.toml`. If you look at the comment there, I think the coverage issue has to do with the lines that get ignored.

caleb531: @eliotwrobson Ohh, I think I know what's going on.

I discovered with one of my recent projects that the `...` exclusion (which was originally added because `@overload` lines weren't getting covered for `DFA._cross_product`) were causing coverage weirdness.

You added the `@overload` decorators back in #131, but they were simplified and ultimately removed in #139. Therefore, it should be safe to remove `...` from the `exclude_lines` array.

P.S. Agreed about moving to `.coveragerc` to `pyproject.toml`, since it makes the coverage config the same format as the other config.

eliotwrobson: @caleb531 I think that did it, removed the exclusion of the `...` here https://github.com/caleb531/automata/pull/129/commits/22317ac972e19b9499fa7dde0891bb6c2f0282a5

caleb531: @eliotwrobson Okay great! Also, regarding the exposing of types via `py.typed`, one way you could test this is to create a new Python project/virtualenv and use `pip install -e` to install your local clone of automata-lib into it. Then open your editor and see if the type checking with mypy is there (will probably need to install mypy as well, and make sure it is enabled in your editor/workspace).

automata/fa/nfa.py
```diff
@@ -1009,3 +970,89 @@ def add_any_transition(start_state_dict, end_state):
...
+        self,
+    ) -> Generator[Tuple[NFAStateT, NFAStateT, str], None, None]:
+        return (
+            (from_, to_, symbol)
+            for from_, lookup in self.transitions.items()
+            for symbol, to_lookup in lookup.items()
+            for to_ in to_lookup
+        )
+
+    def _get_input_path(
```
eliotwrobson, line 984: @khoda81 it looks like this function fails on some inputs (see the `test_nfa.py` file and the changes in the latest commit for details https://github.com/caleb531/automata/pull/129/commits/6f1ec9fd7086e9ba7b5d65eb459ae627ae9cfcea). It also looks like we can't use `functools.cache` in Python 3.8, so it would be great to modify this function to remove that usage and fix whatever bugs we're running into.

caleb531: @eliotwrobson @khoda81 On the note about the caching, `functools.lru_cache(maxsize=None)` seems to be the way to go, since it was added back in Python 3.2, and the docs clearly state it is the same as `functools.cache`:

> Returns the same as lru_cache(maxsize=None)
Source: https://docs.python.org/3/library/functools.html#functools.cache

I will go ahead and make this change, since it should only take 2 minutes.

caleb531: @eliotwrobson @khoda81 Okay, change is made—please remember to pull.

Also, I accidentally resolved the conversation, but I unresolved it now since there is still the matter of behavioral correctness to discuss/resolve.

eliotwrobson:
@caleb531 @khoda81 I've pushed changes https://github.com/caleb531/automata/pull/129/commits/10706a6406d9057a727e2ab49371a3f0b354c009 that fix the correctness issue I was seeing (at least when testing with the small example NFA), along with tests verifying this for that particular instance. There have been a few uncovered lines that have been exposed, but the test cases are checking for most of the important issues related to correctness.

The preceding issue was the last major obstacle to this PR. All that remains is to add a few test cases that ensure correctness and cover the last remaining lines. I should be able to attack some of these tests, but would greatly appreciate some help on this, as there are a few pesky uncovered lines that require some weird setup code to hit. I also want to be sure about the optimality of the new accepting path finding algorithm.

automata/fa/nfa.py
eliotwrobson: Rewrote this algorithm to fix the correctness issue! https://github.com/caleb531/automata/pull/129/commits/10706a6406d9057a727e2ab49371a3f0b354c009

automata/fa/fa.py
eliotwrobson: Added this here https://github.com/caleb531/automata/pull/129/commits/10706a6406d9057a727e2ab49371a3f0b354c009

automata/fa/fa.py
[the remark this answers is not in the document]
eliotwrobson: @caleb531 what are your thoughts about this? I'm not sure if making the directory if it doesn't exist is necessarily the best thing (though I'm not leaning hard either way on this, not sure what best practices are).

automata/fa/fa.py
```diff
@@ -1,19 +1,233 @@
...
+        graph.add_nodes_from(nonfinal_states, shape="circle", fontsize=font_size_str)
+        graph.add_nodes_from(final_states, shape="doublecircle", fontsize=font_size_str)
+
+        is_edge_drawn = defaultdict(lambda: False)
+        if input_str is not None:
+            input_path, is_accepted = self._get_input_path(input_str=input_str)
+
+            start_color = coloraide.Color("#ff0")
+            end_color = (
+                coloraide.Color("#0f0") if is_accepted else coloraide.Color("#f00")
```
eliotwrobson, line 157: Is there any reason to use coloraide specifically? I think the one feature we use is the interpolation, and colours has this while also having more active support (I think).

eliotwrobson: After taking a look, it seems like colour is pretty light on features and doesn't have type annotations set up. I think sticking with coloraide is actually the way to go.

automata/fa/nfa.py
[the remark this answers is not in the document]
eliotwrobson: Resolving since this function was rewritten.

automata/fa/fa.py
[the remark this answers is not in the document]
eliotwrobson: This is now being tested.

caleb531:
@eliotwrobson @khoda81 At this point, I'm tempted to just bypass the slight drop in coverage (to address in a separate PR) for the sake of merging this PR. It will unblock #147 and any other future PRs that touch these files (which would often be the case). I've also been realizing that achieving a perfect 100% code coverage can be impractical or even unattainable, sometimes.

@eliotwrobson I think my only requirement for wrapping up this PR at this point would be to work through the remaining comment threads to reach resolution for all of them.

automata/fa/fa.py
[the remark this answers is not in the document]
caleb531: @eliotwrobson Sorry, never got back to this. I would go with the practice that I see in the Python stdlib and **not** create the directory automatically if it doesn't exist. In the stdlib, the user is always required to use something like `os.makedirs` for that behavior, and we should follow suit.

pyproject.toml
caleb531: @eliotwrobson Just following up—did you ever get a chance to verify the functioning of `py.typed`?

eliotwrobson:
@caleb531 sounds like a plan. I closed all of the conversations related to changes that I already made. There are only 4 remaining ones open, and they're all fairly tangential or something that we can resolve with testing later (minor stuff related to how the project is configured). I'll make another pass.

pyproject.toml
eliotwrobson: I think this is all working now. We'll want to test one more time before a release but I think it should be fine. I'll go ahead and resolve the discussion + delete some of the TODO comments in this file (as they have been resolved).

automata/fa/fa.py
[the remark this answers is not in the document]
eliotwrobson: That makes sense to me, I'll delete the commented out code and resolve.

automata/fa/fa.py
eliotwrobson: Going to go ahead and make this into a separate issue (and resolve) since it's somewhat orthogonal to the goals of this PR, and shouldn't really be THAT much work.

eliotwrobson:
@caleb531 I've resolved all of the open comment threads on this PR. The only dangling items are:

- The coverage, where I agree with you that we should merge and then maybe add a test that everything works without the optional dependencies later.
- The documentation, which I'm inclined to go over later since this PR doesn't really change the API in a substantial way, and we can add some of the new helper functions to the public interface before release.

Other than those, we have very high coverage and tests for most everything that's been added. Once you give the code here a quick lookover / are comfortable, I think we're good to go ahead and merge!

caleb531 (approved):
@eliotwrobson @khoda81 This works for me! I've just approved everything, so I'll go ahead and merge now.

PR #130, opened 2023-03-26
Better numbering usage + other small updates

eliotwrobson:
Uses a new counter each time a regex gets parsed. This is useful because parsing multiple regexes could use a large number of states, and having the complexity of the state name increase too much could slow things down.

Also, small changes to iteration logic (using `zip-repeat` which is faster than `product`), and `isdisjoint` for faster set intersection checking.


caleb531 (approved):
@eliotwrobson This all looks great to me! Thank you for the PR!

Merging now...

PR #131, opened 2023-03-29
[WIP] Type hints

eliotwrobson:
Work in progress branch to introduce type hints. Should probably wait until https://github.com/caleb531/automata/pull/121 and https://github.com/caleb531/automata/pull/129 get merged before doing the rest, since these will alter the type signatures that get used.

Also, need to add a mypy step to the CI workflow to check these annotations.


automata/base/automaton.py
```diff
@@ -95,19 +108,20 @@ def _validate_final_states(self):
...
     @property
-    def input_parameters(self):
+    def input_parameters(self) -> Dict[str, Any]:
         """Return the public attributes for this automaton."""
         return {attr_name: getattr(self, attr_name)
                 for attr_name in self.__slots__
                 if not attr_name.startswith('_')}
 
-    def copy(self):
+    #TODO add annotation to here?
```
caleb531, line 117: @eliotwrobson I would think a return type isn't needed here, but really depends on whether Python can infer the return type from the below call to `__init__` (this is something TypeScript does pretty well, but not sure how Python type-checkers handle this).

In other words, how good is mypy's or Pyright type inferencing ability? Considering that Pyright is developed by Microsoft (the same company that brought us TypeScript), I would think it would be pretty capable with inferencing.

automata/base/automaton.py
```diff
@@ -8,12 +8,24 @@
 import automata.base.config as global_config
 import automata.base.exceptions as exceptions
 from automata.base.utils import freeze_value
+from typing import Any, Mapping, NoReturn, AbstractSet, Tuple, Dict, Generator
 
+AutomatonStateT = Any
+PathT = Mapping[str, Any]
```
caleb531, line 14: @eliotwrobson I'm guessing the fact that you've set the type for the path value as `Any` is part of solution to solving the Liskov problem (at least from a type perspective)? I was thinking about this earlier, and that's exactly what I was going to suggest doing (at least for the abstract base classes).

I also see you're getting more specific at the subclass level as well (e.g. with `NFATransitionsT`), which was exactly what I was envisioning. So thanks for that! 😀

I know you also mentioned some sort of wrapper to solve the Liskov problem—curious to hear more about what you had in mind with that.

automata/base/automaton.py
```diff
@@ -22,39 +34,40 @@ def __init__(self, **kwargs):
...
         """Return the object's state, described by its input parameters"""
         return self.input_parameters
 
-    def __setstate__(self, d):
+    def __setstate__(self, d: Dict[str, Any]) -> None:
         """Restore the object state from its input parameters"""
         # Notice that the default __setstate__ method won't work
         #   because __setattr__ is disabled due to immutability
-        self.__init__(**d)
+        self.__init__(**d) # type: ignore
```
caleb531, line 62: @eliotwrobson What is the purpose of the ignore for this particular line?

automata/fa/dfa.py
```diff
@@ -644,7 +695,7 @@ def isfinite(self):
         except exceptions.EmptyLanguageException:
             return True
 
-    def random_word(self, k, *, seed=None):
+    def random_word(self, k: int, *, seed:Optional[int]=None) -> str:
```
caleb531, line 698: @eliotwrobson Is the space after the colon normally absent when a default value is involved? I noticed that the `predecessor` function still included the space on line 723 (`strict: bool=True`).

automata/base/automaton.py
```diff
@@ -8,12 +8,24 @@
 import automata.base.config as global_config
 import automata.base.exceptions as exceptions
 from automata.base.utils import freeze_value
+from typing import Any, Mapping, NoReturn, AbstractSet, Tuple, Dict, Generator
 
+AutomatonStateT = Any
```
caleb531, line 13: @eliotwrobson So I remember when I was taking a stab on my own at adding types to this library, I was running into an issue where, because we were coercing most `__init__` input parameters to be immutable structures, the types of the `__init__` inputs wouldn't necessarily be the same as what the automata is storing.

For example, you could pass `List[Hashable]` to the `states` parameter on init, but that structure gets converted to something that looks more like `Set[Hashable]`, and the latter is what's actually stored on the instance. Another consequence of this was that I had to use `Hashable` in most places where `Any` would otherwise be used.

Do you have any thoughts on how to handle these complexities?

caleb531:
@eliotwrobson A few thoughts and questions:

## New merge conflict

It looks like the recent merge (from maybe #121) introduced a merge conflict in the nfa.py module. Just wanted to let you know

## Type name conventions

Is suffixing type variables with `T` an appropriate convention for custom Python types? I do recognize that some languages (like C#) have conventions like the `I` prefix for interface names, but wasn't sure what is the case with Python types.

I mean, a `T` suffix would certainly make sense for custom types, because (for example) `NFAStateT` would look too much like an actual class.

## Type annotations for everything else?

Your PR looks great so far, and thanks for adding type annotations to the PDAConfiguration and PDAStack classes. But if we're gonna embrace types in this library, _every_ class/module should have type annotations.

How easy would it be to add types to the remaining `DPDA`, `NPDA`, `DTM`, `NTM`, and `MNTM` classes? It's perfectly okay to say "that's too much work / out of scope for me" or something—I can do them at the end of the day.

I guess I just ask because I've barely worked with types in Python and don't know that side of the stdlib very well, so you would probably be a lot faster than me at adding those annotations. But at the end of the day, I don't want to pressure you to do more than you're prefer to do.

## Other things

I left you miscellaneous other comments in the PR, mostly small observations or questions I had around design desisions. Please review when you have time and let me know your thoughts!

eliotwrobson:
@caleb531

> Is suffixing type variables with `T` an appropriate convention for custom Python types? I do recognize that some languages (like C#) have conventions like the `I` prefix for interface names, but wasn't sure what is the case with Python types.

I'm actually not sure, from the other type annotations I've seen, there's not really a standard convention, I just do this sometimes to make it easier to tell what's a type and what's a class. But it's no trouble to change to/from this convention depending on what you prefer. Maybe there's a convention documented somewhere?

> Your PR looks great so far, and thanks for adding type annotations to the PDAConfiguration and PDAStack classes. But if we're gonna embrace types in this library, _every_ class/module should have type annotations.

Definitely agree, I only stopped at the NFA/DFA because I ran out of time the day I was doing this but wanted to put up the work in progress as soon as possible.

> How easy would it be to add types to the remaining `DPDA`, `NPDA`, `DTM`, `NTM`, and `MNTM` classes? It's perfectly okay to say "that's too much work / out of scope for me" or something—I can do them at the end of the day.

Actually not hard, the most difficult files to annotate are the NFA/DFA because these are the longest.

> I guess I just ask because I've barely worked with types in Python and don't know that side of the stdlib very well, so you would probably be a lot faster than me at adding those annotations. But at the end of the day, I don't want to pressure you to do more than you're prefer to do.

I'm happy to! (and intend on doing so as part of this PR before merging). But if you'd like to take a stab at it, you're certainly welcome to. If you want to work with types some in Python, I actually think adding annotations to a library like this is a good way to learn.

automata/fa/dfa.py
eliotwrobson: Oh the space after the colon is usually present, I was just doing this in a hurry and intended on running an autoformatter before merging.

automata/base/automaton.py
eliotwrobson: This is the error that gets raised:

image

Not quite sure what this particular function is doing, so it may make sense to do a minor refactor to avoid this unsound access.

caleb531: @eliotwrobson Defining the `__getstate__` and `__setstate__` methods enable the automaton objects to be picked—this ability was added back in #125, and is one I'd like to keep (especially since it was a requested feature).

After reviewing that PR, it seems the pickling capability only really applies to mutable automaton instances anyway (i.e. with `global_config.allow_mutable_automata = true`). So I don't think we can use `self.__class__(...)` like the `copy` method does, since that creates a new instance.

automata/fa/dfa.py
caleb531: @eliotwrobson Got it, will resolve this thread then, especially since the autoformatter will take care of these things.

automata/base/automaton.py
eliotwrobson: So with type annotations, the inferencing is actually pretty good in Pyright (which does a great job at powering Pylance, IMO), but the issue has to do with the fact that Python typechecking is very permissive by design. So even though Pylance will correctly infer `copy` returns the same type, other code that is typechecked won't strictly enforce this inference. 

Here's a concrete example: Without an explicit annotation, if client code calls the `copy` function and then called a method that did not exist, this would be incompatible with the inferred return type from `copy`, but it would be compatible with `Any` (as this doesn't contradict the existing blank annotation). This behavior also isn't uniform across typecheckers.

All of this is to say that Python typechecking is meant to be as flexible as possible, but this comes at the cost of being very permissive. I think it's generally a good idea to be as explicit as possible to get as much correctness out of the typechecker as we can.

caleb531: @eliotwrobson Well, as the Zen of Python states, "Explicit is better than implicit". This is the Way. 😉

So let's leave the TODO, and add the return type annotation for explicitness. I believe I should have access to commit to your branch, so I can update this now.

automata/base/automaton.py
eliotwrobson: That makes sense, having the ignore here isn't a huge deal and I agree that pickling support is a feature worth keeping.

automata/base/automaton.py
eliotwrobson: Yes, the `Any` here is to allow the derived classes to override this. The Liskov issue actually wasn't with this particular type, but it came up when annotating some validation function in one of the TM classes or something.

automata/base/automaton.py
```diff
@@ -8,7 +8,7 @@
 import automata.base.config as global_config
 import automata.base.exceptions as exceptions
 from automata.base.utils import freeze_value
-from typing import Any, Mapping, NoReturn, AbstractSet, Tuple, Dict, Generator
+from typing import Any, Mapping, NoReturn, AbstractSet, Tuple, Dict, Generator, Self
```
eliotwrobson: @caleb531 this is 3.11 only, you need to do `from typing_extensions import Self`.

caleb531: @eliotwrobson Ha yep, was just figuring that out after I noticed the failing build. Glad to know it's in `typing_extensions`. Thanks for the tip.

caleb531:
@eliotwrobson Just a heads up—force-pushed a correction to the import order for the `Self` fix. Just wanted to let you know in case you already pulled my original (now lost-in-space) commit.

eliotwrobson:
@caleb531 are you able to run `mypy .` locally without any errors being thrown? It seems to have issues with the `Self` type, but I think this might just be my machine for some reason.

caleb531:
> are you able to run mypy . locally without any errors being thrown? It seems to have issues with the Self type, but I think this might just be my machine for some reason.

@eliotwrobson No issues on my end, apparently.

Screenshot 2023-03-30 at 7 16 53 PM

```
> mypy --version
mypy 1.1.1 (compiled: yes)
```

eliotwrobson:
@caleb531 ahhh, I think I have a really old mypy version, that must be the issue.

Forgot to mention in a previous comment: The next item related to this PR for me is going to be adding annotations to the regex library. Those are a little on the sophisticated side and might take me a while, so if you want to try your hand at annotating the other classes (TM, DTM, etc.) you're welcome to commit directly to my branch. I think that things will go much faster that way 😃

eliotwrobson:
@caleb531 Will go ahead and rebase, should be pretty simple.

With respect to removing flake8-isort, here's a guide on how to set up black with other tools: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html

Basically, you need to set some minor configuration options, but flake8 plays really nice with black if you set up the plugin.

.github/workflows/format.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          sudo apt-get install graphviz
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Format Python files with black
+        uses: psf/black@stable
```
eliotwrobson: @caleb531 is this intended to check formatting on commit? If so, you should be using `flake8` with the `flake8-black` and `flake8-isort` plugins. Just running black without flake8 won't check for things like unused imports, for example.

caleb531: @eliotwrobson Probably a mistake due to my unfamiliarity with black. I've only ever used flake8+pycodestyle for the longest time in my projects, and thought that pycodestyle (a required dependency of flake8) would conflict with black in this case, hence the removal of flake8 and keep just pyflakes.

At any rate, since we haven't formatted the other files in the project yet, I reverted the addition of the CI action for now so I don't keep getting 'failed job' emails on every commit due to the formatter complaining about those files.

eliotwrobson: Pycodestyle shouldn't conflict with black. It's a bit tricky, but I've actually set up something similar in another project. Basically, you need the following dependencies:
```
flake8-black
flake8-isort
Flake8-pyproject
flake8
black
isort
```
and then in `pyproject.toml` you need the following configurations
```
[tool.flake8]
# Black compatibility
max-line-length = 88
extend-ignore = ["E203", "W503"]

[tool.isort]
profile = "black"
```
This will essentially have `flake8` run `isort` and `black` on its own and check for inconsistencies with the way those tools do formatting. Then, you can have the workflows just run `flake8 .`. I would suggest having the formatting run before/after testing on the existing CI jobs, since this means not having to wait for setup to run twice for each Python version.

caleb531: >  I would suggest having the formatting run before/after testing on the existing CI jobs, since this means not having to wait for setup to run twice for each Python version.

I'd prefer to have a separate GitHub action because if the `tests` action fails, I want to assume without issue that a test failed/errored, rather than just a formatting issue or linter error creeping in.

> Like I said above, I think this is best done in a different PR, where we can run the autoformatting over everything as well.

What do you mean? Should I disable the format-on-file-save behavior so that the other files don't get auto-formatted? I thought you were going to run everything through black (i.e. format everything) anyway at the end of this PR. If that's the case, then what would be the purpose of the second PR?

caleb531: @eliotwrobson 

>  I would suggest having the formatting run before/after testing on the existing CI jobs, since this means not having to wait for setup to run twice for each Python version.

I'd prefer to have a separate GitHub action because if the `tests` action fails, I want to assume without issue that a test failed/errored, rather than just a formatting issue or linter error creeping in.

> Like I said above, I think this is best done in a different PR, where we can run the autoformatting over everything as well.

What do you mean? Should I disable the format-on-file-save behavior so that the other files don't get auto-formatted? I thought you were going to run everything through black (i.e. format everything) anyway at the end of this PR. If that's the case, then what would be the purpose of the second PR?

eliotwrobson: @caleb531 

> What do you mean? Should I disable the format-on-file-save behavior so that the other files don't get auto-formatted? I thought you were going to run everything through black (i.e. format everything) anyway at the end of this PR. If that's the case, then what would be the purpose of the second PR?

Actually, on second thought I think it's fine. My thinking is that running the autoformatting right now is probably not the best idea since if something like #129 gets merged before this, it will be difficult to deal with merge conflicts because the base branch will have non-auto-formatted code. But if we just wait to run then it should be alright.

caleb531: @eliotwrobson I see, that makes sense. I actually already ran black on automaton.py in a recent commit (when I added the VS Code configuration to format on save), but I'm pushing a commit now to disable that until a later point.

caleb531:
@eliotwrobson A change I'm thinking of making in this PR, for better abstraction: instead of using the type `str` everywhere to represent the type of automaton symbols, let's assign that to a variable and use that variable everywhere. e.g.:

```python
AutomatonSymbolT = str
```

This would make the vast majority of symbol-related type annotations a bit more agnostic to the specific type of symbols currently supported. So in the future, if we support another type of symbol, it's easy to extend all existing type annotations with a one-line change. I would also be happy to make this change, as long as I'm not stepping on your toes. Thoughts?

eliotwrobson:
@caleb531 sounds like a great idea! I'm busy for the rest of today at least so feel free to go right ahead.

caleb531:
@eliotwrobson On second thought, not going to change the type for now because it's gonna get confusing/tricky to define higher-level semantics for everything represented by `str` in the library. For example, would the type for `input_str` change? Would we also be creating a new type to represent a word (e.g. `AutomatonWordT`)? 

And then what about things like the regex parser? Looking through the parser code, I see many places where `str` is used as a type (as you've committed recently), but I'm not sure I can tell when `str` refers to a word vs. a symbol.

caleb531:
@eliotwrobson Also, I've finished merging in my formatting changes into your branch. I opted to do a merge instead of a rebase because it made the conflict resolution a little easier, but I tried my best to preserve all the type annotations you've added so far.

eliotwrobson:
@caleb531 sounds good! I think it's fine to wait until we get a first round of type annotations before doing something more intricate.

automata/pda/stack.py
```diff
@@ -1,24 +1,29 @@
...
 import collections
+from typing import Iterator, Tuple, Union
 
 
 class PDAStack(collections.namedtuple("PDAStack", ["stack"])):
     """A PDA stack."""
 
+    stack: Union[Tuple[str], str]
+
+    # TODO can we get rid of this? Kinda weird inheritance here
```
caleb531, line 14: @eliotwrobson Just saw this question and wanted to share my thoughts:

## TL;DR

I'm fine with it, kinda like it, and would prefer to keep it lol.

## Long Answer

The inheritance from namedtuple was contributed by @YtvwlD back in 4e645f16fae5fdc6d11f6a8ef516d5f02bda3281 as part of the NPDA implementation (which he also kindly contributed in full). As I understand it, the inheritance from namedtuple provides automatic immutability and hashability characteristics without much code, which was helpful when implementing the nondeterminism in NPDA (and NTM as well).

I think this approach is fine for auxiliary library primitives like PDAStack that are useful in many contexts. Rather than creating a wrapper around a builtin primitive (which requires initializing internal state, duplicating operator overloading, etc.), we just inherit from that builtin primitive and automatically gain all of its characteristics and behavior. I also suspect there's a value-add in terms of efficiency, because the nondeterministic automata types can generate a great many permutations of configurations.

eliotwrobson: That makes sense, thinking about why I left this comment, the inheritance isn't such a big deal, but the use of the `__new__` seems to cause issues with typechecking. I think it's not a big deal to ignore it if we want to keep this data structure the way it is, but it's worth re-evaluating (at least if we can get rid of the `__new__` call).

caleb531: @eliotwrobson I see. I'm fine to re-evaluate, although I haven't yet run into any issues with the type-checking related to __new__ (I've finished typing the *PDA modules already, and mypy isn't complaining so far). But given that continued evolution is in the spirit of this library's development, I'd be fine to re-evaluate later on.

With that, I will keep the TODO comment so we can come back to it at a later time.

automata/pda/pda.py
```diff
@@ -2,16 +2,33 @@
...
 import automata.base.exceptions as exceptions
 import automata.pda.exceptions as pda_exceptions
-from automata.base.automaton import Automaton
+from automata.base.automaton import Automaton, AutomatonStateT, AutomatonTransitionsT
+from automata.pda.configuration import PDAConfiguration
+from automata.pda.stack import PDAStack
+
+PDAStateT = AutomatonStateT
+PDATransitionsT = AutomatonTransitionsT
+PDAAcceptanceModeT = Union[
```
eliotwrobson, line 17: @caleb531 
```suggestion
PDAAcceptanceModeT = Literal["final_state", "empty_stack", "both"]
```

caleb531: @eliotwrobson You can do that?? I mean, uhh, I was going to fix that eventually lol. 😅

Ha and you can see how my inexperience with Python types is starting to show. So thanks for the pointer!

eliotwrobson:
@caleb531 now that the annotations are getting more complete, I think it's a good opportunity to start setting up type checking tests. I'm not sure how this is typically done for packages (not many packages actually have inline type hints), but I think just running mypy before running tests should be enough on CI.

To make the type stubs visible externally, I think we need to add a file as in this stack overflow post: https://stackoverflow.com/questions/62979389/is-there-a-best-practice-to-make-a-package-pep-561-compliant

caleb531:
@eliotwrobson Hm, I was looking at [flake8-mypy](https://pypi.org/project/flake8-mypy/) as a potential option (so that we can have all static code analysis managed through flake8), but it seems there may be some version compatibility issues (or may not be, one of us just needs to take a closer look). But if mypy needs to be run separately, I would prefer it be part of the [`lint`](https://github.com/caleb531/automata/blob/develop/.github/workflows/lint.yml) action, rather than `tests`.

I have a high-priority bug that I need to work on fixing for one of my other projects, so I don't have time to look into this at the moment. But let me know your thoughts, and I'll be happy to review anything you contribute further!

eliotwrobson:
@caleb531 no worries! At some point I'll try and take a look at some other libraries with integrated types and see how they're configured. Though It looks like flake8-mypy hasn't been maintained for some time. I'm not exactly sure what the preferred configuration is for modern projects.

eliotwrobson:
@caleb531 in the latest push, I've added a `py.typed` file as described here: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

I didn't add type annotations to test cases, but I was wondering what else you'd like to see done before this gets merged. The changes here are pretty extensive, so I think it might make sense to do follow-up typing in the test files so that other branches don't get too out of sync with this one.

automata/pda/stack.py
eliotwrobson: @caleb531 looking at this again, I realized what was going on. When you try to have this class inherit from `NamedTuple` from types instead of collections, Python doesn't allow you to override the `__new__` function, which is what this comment was referring to initially. If we find a way to avoid the override of `__new__`, then this (and other) classes can use the cleaner style of inheritance for named tuples.

caleb531: @eliotwrobson I see what you mean now. [The new-style inheritance syntax for named tuples](http://zecong.hu/2019/08/10/inheritance-for-namedtuples/) _is_ much cleaner. But yes, if we want to inherit from named tuple at all, the `__new__` is still needed because we need to accept an input argument that are cast to something else.

Another option that I was playing around with (though I have mixed feelings about) is using a dataclass (and just forget about inheriting from named tuple in the first place). I still don't know about the performance differences with inheriting from a native type, and I am not a huge fan of the imperative casting of the `stack` attribute after-the-fact, but I'm just trying to think creatively. Here's a small proof-of-concept I wrote.

```python
#!/usr/bin/env python3

from dataclasses import dataclass
from typing import Tuple


@dataclass(frozen=True)
class PDAStack:

    stack: Tuple[str, ...]

    def __post_init__(self):
        object.__setattr__(self, 'stack', tuple(self.stack))


def main():
    stack = PDAStack(['a', 'b', 'c'])
    print(stack)  # PDAStack(stack=('a', 'b', 'c'))


if __name__ == '__main__':
    main()
````

caleb531:
@eliotwrobson I agree that I don't want this PR to sit for too long, especially given the other PRs we want to wrap up. I think we are in a place where we can merge this one safely, and then follow up with additional type PRs (e.g. for better mypy integration, types in tests, etc.). But as I see it, because mypy isn't currently complaining about the lack of types in tests, I think it's fine to prepare this PR for final review / merge.

automata/pda/stack.py
caleb531: @eliotwrobson Given my comment on the main thread just a moment ago, we can revisit this in a future PR/thread, especially since we have the TODO in place to remind us. So I think I'll resolve this comment thread now so we can wrap up this PR.

automata/base/automaton.py
caleb531: @eliotwrobson Resolving this thread, as I've been using `AbstractSet` per your recommendation and have been pretty happy / haven't had any issues.

eliotwrobson:
@caleb531 agreed. Since this is going into the develop branch I think it's fine to table some of the discussion items since they were pretty minor anyway. My only major item was a few missing type annotations but that can be handled in a followup.

caleb531 (approved):

PR #133, opened 2023-04-04
Simplify minify logic

EduardoGoulart1:
@caleb531 @eliotwrobson I wanted to drop a proposal to decouple the logic of the cross-product. The idea is to turn the exploration method into a function that receives an initial state and an expansion function and applies BFS using only that. This makes the current implementation cleaner and will also significantly simplify the implementation for `allow_partial`

I did some microbenchmarks and noted that the overhead is relatively small.

Please do not make a full review of this PR. If you approve the design I will clean it up to make it ready to merge.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -391,22 +391,14 @@ class DFA(fa.FA):
             retain_names=retain_names
         )
 
-        if minify:
-            return self._minify(
-                reachable_states=new_states,
-                input_symbols=self.input_symbols,
-                transitions=new_transitions,
-                initial_state=new_initial_state,
-                reachable_final_states=new_final_states,
-                retain_names=retain_names)
-
-        return self.__class__(
+        ret = self.__class__(
             states=new_states,
             input_symbols=self.input_symbols,
             transitions=new_transitions,
             initial_state=new_initial_state,
             final_states=new_final_states
         )
+        return ret.minify(retain_names=retain_names) if minify else ret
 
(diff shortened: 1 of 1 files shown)


EduardoGoulart1:
By using this design, support for allow_partial can be added quite naturally by adapting the state transition functions.

Also, the issubset, isdisjoint and isempty can be implemented faster by switching to DFS instead of BFS

eliotwrobson:
This design seems cool! I think we might want to wait till an initial version of the type hints PR is merged, since this changes a significant number of those. But I think this refactor with the generator makes things more flexible.

> Also, the `issubset`, `isdisjoint` and `isempty` methods can be implemented faster by switching to DFS instead of BFS

I don't think this is always true. Those functions are searching for a single state as a counterexample, in the worst case they all have to search every reachable state.

caleb531:
@EduardoGoulart1 I like the approach of this architecture, but also agree with @eliotwrobson about the BFS matter. Fortunately, I have just merged the types PR (#131) so you can continue to work on this PR.

EduardoGoulart1:
Sorry, I think I expressed myself badly. Yes, both have the same worst-case complexity. But I was starting from the assumption that end states are "deeper" in the DFA, which we might want to verify.

In any case, `isempty` will profit from early-stopping. But, I won't change anything related to it in this PR

EduardoGoulart1:
Closing this one. See https://github.com/caleb531/automata/pull/139

PR #136, opened 2023-04-05
Regex empty string

### LINKED ISSUE #128: How to express the empty string in regexes?
Is there a way to specify the empty string as part of regexes? For example, Python's built-in re.compile allows expressions like `'()'` or `'|'`, but the parser in this library throws an error on both.

eliotwrobson:
Added empty string to regex with `()`. Resolves #128.


caleb531 (approved):
This looks great to me! 👌

Issue #135, opened 2023-04-04
`validate` return type


eliotwrobson:
@caleb531 looking at the code again, I observed that the `validate` function either raises an exception (in the fail case) or returns `True` (if validation succeeded). I think that the return type should be omitted, since no exception being raised informs the client code that validation has succeeded, and the existing return value can always be discarded. I think the situation is similar to that described here: https://stackoverflow.com/questions/61304761/is-it-good-practice-to-return-true-when-success-otherwise-raise-exception-in-pyt

caleb531:
@eliotwrobson 

## To return or to have side effects?

When you say "the return type should be omitted", it sounds like you mean to rid of the `return True` altogether and have the function either (implicitly) return `None` _or_ raise an exception. If that is what you mean (as opposed to merely omitting the annotation), that does sense to me, since I am have tried to employ the following principle in my own development:

> A function should either return a value, or have side effects, but not both.

Now, in the context of Python, I do not consider raising an exception to be a side effect that breaks this rule, as it is a common pattern used in the Python stdlib and elsewhere.

## Implicitly returning `None` / no return statement makes sense

However, I do agree that returning a constant value like `True` feels a bit useless. I think it was a decision I made when I originally wrote the library, and I suppose it just stuck around all these years. And while it does make testing a little easier (there is no `assertNotRaises` assertion in Python), there are decent alternatives (such as checking if the return type is implicitly `None`, if we are okay with the tests caring about the return value in the first place).

If we implicitly return `None`, it _would_ technically be a breaking change, albeit a minor one. The only case I can think of where that could break someone's code is if they are calling `validate()` as the condition of an `if` block, and therefore are relying on a truthy value being returned for the condition to pass. But they probably shouldn't be doing that anyway, since in the case of an invalid machine, an exception is raised rather than `False` being returned. So the correct usage would only ever be calling the method in a `try..except` and without capturing the return value.

## Public or private?

And with the recent immutability changes, `validate()` has (in many ways) effectively become more of a private method (although technically still not, since we added a global option to re-enable mutability, so allowing the user to call the method may still be useful).

eliotwrobson:
@caleb531 yes, you phrased it better than I did. Effectively this would just be removing the `return True` line at the end of each validate function. And while it would _technically_ be a breaking change, code relying on that behavior would need to be in a `try..except` block to work correctly.

caleb531:
@eliotwrobson Agreed. And closing this issue now that #137 has been merged.

PR #137, opened 2023-04-05
Validate type change

eliotwrobson:
See title, resolves https://github.com/caleb531/automata/issues/135

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -266,14 +266,13 @@ class DFA(fa.FA):
         self._validate_transition_invalid_symbols(start_state, paths)
         self._validate_transition_end_states(start_state, paths)
 
-    def validate(self):
+    def validate(self) -> None:
         """Return True if this DFA is internally consistent."""
         self._validate_transition_start_states()
         for start_state, paths in self.transitions.items():
             self._validate_transitions(start_state, paths)
         self._validate_initial_state()
         self._validate_final_states()
-        return True
 
     def _get_next_current_state(
         self, current_state: DFAStateT, input_symbol: str
--- a/tests/test_regex.py
+++ b/tests/test_regex.py
@@ -15,11 +15,8 @@ class TestRegex(unittest.TestCase):
 
     def test_validate_valid(self):
         """Should pass validation for valid regular expression"""
-        try:
-            re.validate("a*")
-            re.validate("b|a?*")
-        except Exception:
-            self.fail("Regex validation raised an exception when not expected to.")
+        re.validate("a*")
+        re.validate("b|a?*")
 
     def test_validate_invalid(self):
         """Should raise error for invalid regular expressions"""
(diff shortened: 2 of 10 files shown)


automata/fa/dfa.py
```diff
@@ -266,14 +266,13 @@ def _validate_transitions(self, start_state: DFAStateT, paths: DFAPathT) -> None
         self._validate_transition_invalid_symbols(start_state, paths)
         self._validate_transition_end_states(start_state, paths)
 
-    def validate(self):
+    def validate(self) -> None:
```
caleb531, line 269: @eliotwrobson Do we need to explicitly declare the return type for the `validate` methods on the subclasses? I would think the return type would be inherited from the superclass (`Automaton`, in this case).

tests/test_regex.py
```diff
@@ -15,8 +15,11 @@ class TestRegex(unittest.TestCase):
 
     def test_validate_valid(self):
         """Should pass validation for valid regular expression"""
-        self.assertTrue(re.validate("a*"))
-        self.assertTrue(re.validate("b|a?*"))
+        try:
+            re.validate("a*")
+            re.validate("b|a?*")
+        except Exception:
+            self.assertTrue(False)
```
caleb531, line 22: @eliotwrobson Would be better to use `self.fail()` with an appropriate error message here. See:
https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail

caleb531: @eliotwrobson On second thought, do we even need a `try..except` here? An exception raised during a test already counts as a failure.

automata/fa/dfa.py
eliotwrobson: I don't think this is technically necessary, but I think we should strive to have annotations on as many functions as possible (not sure if there's precedent for that anywhere but typecheckers include flags that disallow untyped functions).

tests/test_regex.py
eliotwrobson: That is true, I added the fail statement so I suppose having the explicit error message is better. To be honest I thought about this but just defaulted to the more explicit check.

caleb531: @eliotwrobson Yes, but you're swallowing up the exception that was raised. From a debugging perspective, that would be annoying because you won't know what caused the failure. Was it a `TypeError`, `ValueError`, or something else?

Thinking it over, I would ask that the `try..catch` be removed. For me, practicality beats purity in this case, and I would much rather let any exceptions in my tests be raised naturally (so I have access to the full stack trace and all that good stuff).

eliotwrobson: That makes sense, will go ahead and remove

eliotwrobson: @caleb531 done here: https://github.com/caleb531/automata/pull/137/commits/d21a975d96ad4f0daf79e6f6f0b1e9e3c07abacb

PR #139, opened 2023-04-05
Simplified binary operations for DFAs

EduardoGoulart1:
This is the PR. Hope it is good to go now

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from collections import defaultdict, deque
 from itertools import chain, count
 from random import Random
-from typing import (
+from typing import (  # noqa Fixes a false positive where flake8 thinks that Literal is imported/unused even though it's not imported at all here
     AbstractSet,
     Any,
     Callable,
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -61,9 +61,7 @@ class DFA(fa.FA):
     _word_cache: List[DefaultDict[DFAStateT, List[str]]]
     _count_cache: List[DefaultDict[DFAStateT, int]]
 
-    ExpandStateFn = Callable[
-        [DFAStateT], Generator[Tuple[DFASymbolT, DFAStateT], None, None]
-    ]
+    ExpandStateFn = Callable[[DFAStateT], Iterator[Tuple[DFASymbolT, DFAStateT]]]
     IsFinalStateFn = Callable[[DFAStateT], bool]
     TargetStateFn = Callable[[DFAStateT], bool]
 
@@ -341,7 +339,7 @@ class DFA(fa.FA):
 
         # Compute reachable states and final states
         bfs_states = DFA._bfs_states(
-            self.initial_state, lambda state: self.transitions[state].items()
+            self.initial_state, lambda state: iter(self.transitions[state].items())
         )
         reachable_states = {*bfs_states}
         reachable_final_states = self.final_states & reachable_states
@@ -548,7 +546,7 @@ class DFA(fa.FA):
 
         if minify:
             bfs_states = DFA._bfs_states(
-                self.initial_state, lambda state: self.transitions[state].items()
+                self.initial_state, lambda state: iter(self.transitions[state].items())
             )
             reachable_states = {*bfs_states}
             reachable_final_states = self.final_states & reachable_states
@@ -575,7 +573,7 @@ class DFA(fa.FA):
     def _bfs_edges(
         initial_state: DFAStateT,
         expand_state_fn: ExpandStateFn,
-    ) -> Generator[Tuple[DFAStateT, DFASymbolT, DFAStateT], None, None]:
+    ) -> Iterator[Tuple[DFAStateT, DFASymbolT, DFAStateT]]:
         """
         Emits the edges (src_state, label, tgt_state) visited by BFS from the
         initial_state. Computes subsequent states using the function expand_state_fn.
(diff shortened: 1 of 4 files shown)


automata/fa/dfa.py
```diff
@@ -63,6 +61,10 @@ class DFA(fa.FA):
     _word_cache: List[DefaultDict[DFAStateT, List[str]]]
     _count_cache: List[DefaultDict[DFAStateT, int]]
 
+    ExpandStateFn = Callable[[DFAStateT], Iterable[Tuple[DFASymbolT, DFAStateT]]]
```
EduardoGoulart1, line 64: I didn't know if there was a better place to add it

EduardoGoulart1:
See https://github.com/caleb531/automata/pull/133 for pass discussion

eliotwrobson:
I haven't had time to do a full review, but everything here looks really good (+1 for accurate type hints everywhere) and cleaner than the previous code. None of the test cases or parts of the public interface are modified, so I'm inclined to say this should be merged soon.

automata/fa/dfa.py
```diff
@@ -652,119 +569,146 @@ def complement(self, *, retain_names: bool = False, minify: bool = True) -> DFA:
...
-        retain_names: bool = False,
-    ) -> Union[bool, DFADataTupleT]:
+                if tgt_state not in visited_set:
+                    visited_set.add(tgt_state)
+                    queue.append(tgt_state)
+
+    @staticmethod
+    def _bfs_states(
+        initial_state: DFAStateT, expand_state_fn: ExpandStateFn
+    ) -> Iterable[DFAStateT]:
```
eliotwrobson, line 597: I think it's technically more precise to annotate this as `Generator[DFAStateT, None, None` instead of just an iterable, since an iterable could be just a list or something.

automata/fa/dfa.py
```diff
@@ -652,119 +569,146 @@ def complement(self, *, retain_names: bool = False, minify: bool = True) -> DFA:
             allow_partial=self.allow_partial,
         )
 
-    # visited_set, product_transitions, product_initial_state_name, final_states
-    DFADataTupleT = Tuple[Set[DFAStateT], DFATransitionsT, DFAStateT, Set[DFAStateT]]
+    @staticmethod
+    def _bfs_edges(
+        initial_state: DFAStateT,
+        expand_state_fn: ExpandStateFn,
+    ) -> Iterable[Tuple[DFAStateT, DFASymbolT, DFAStateT]]:
```
eliotwrobson, line 576: Same as other comment, I think it's better to use a `Generator` annotation.

automata/fa/dfa.py
```diff
@@ -509,34 +480,15 @@ def intersection_function(state_pair: Tuple[DFAStateT, DFAStateT]) -> bool:
...
-        (
-            new_states,
-            new_transitions,
-            new_initial_state,
-            new_final_states,
-        ) = self._cross_product(
-            other,
+        initial_state, expand_state_fn = DFA._cross_product(self, other)
+
+        return DFA._expand_dfa(
```
eliotwrobson, line 485: I think some of these should be `self.__class__` to play nicer with inheritance? Double check with @caleb531

eliotwrobson (approved):
Very minor comments. Nothing major.

automata/fa/dfa.py
caleb531: @EduardoGoulart1 Agreed with @eliotwrobson, because the direct return value of this function is technically a generator object.

automata/fa/dfa.py
caleb531: @EduardoGoulart1 Yes, I generally prefer `self.__class__` (except under rare circumstances) when returning new instances within a class.

automata/fa/dfa.py
caleb531: @eliotwrobson Also noting that I agree `Generator` should be used as the return type.

caleb531: @EduardoGoulart1 Also noting that I agree Generator should be used as the return type.

automata/fa/dfa.py
EduardoGoulart1: I switched to `Iterator` because this is actually what the interface expects. A generator is also an iterator. But if the function would return lists or iterators it would work just fine (in fact, this is what the lambdas in lines 342, 549 and 744 do).

EduardoGoulart1: But not sure how to fix the current link error. Because it complains that I am importing something which I am not :question:

caleb531: @EduardoGoulart1 Seems like it might be a flake8 bug. I've committed a `# noqa` command to your branch, so CI should pass now.

caleb531: @EduardoGoulart1 I also pushed a commit with the `Generator` tweak for `_bfs_edges`, so I think we can mark this thread as resolved now. cc @eliotwrobson

automata/fa/dfa.py
caleb531: I see the code has been updated to use `self.__class__`, so resolving this conversation thread now.

automata/fa/dfa.py
caleb531: Committed this tweak myself, so marking this thread as resolved.

automata/fa/dfa.py
caleb531: @EduardoGoulart1 Decided to move these `*Fn` types to the top-level `dfa` module scope, since that is where the other DFA-specific types currently live.

caleb531:
I've resolved the remaining conversation threads by just committing those small tweaks myself. @EduardoGoulart1 did you have any other changes you wanted to make to this PR before I perform a final review / merge?

cc @eliotwrobson

EduardoGoulart1:
> I've resolved the remaining conversation threads by just committing those small tweaks myself. @EduardoGoulart1 did you have any other changes you wanted to make to this PR before I perform a final review / merge?
> 
> cc @eliotwrobson

No... Go ahead!

PR #141, opened 2023-04-11
Use Self type where applicable

caleb531:
@eliotwrobson Just wanted to get your thoughts on this—I noticed that we still use explicit class types as return types in many places (e.g. `-> DFA` and `-> NFA`). But if we have access to the `Self` and `Type[Self]` types, then these should not be needed, correct?

Per my commit note in this PR, using `Self` where applicable should also eliminate those `TypeVar` types (i.e. `DFAType`), which is actually one of the reasons why `Self` was introduced (see [the "Use in Classmethod Signatures" section in PEP 673](https://peps.python.org/pep-0673/#use-in-classmethod-signatures)).

Anyway, please let me know what you think of this!

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -29,6 +29,7 @@ from typing import (  # noqa Fixes a false positive where flake8 thinks that Lit
 
 import networkx as nx
 from pydot import Dot, Edge, Node
+from typing_extensions import Self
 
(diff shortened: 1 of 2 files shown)


eliotwrobson (approved):
Lgtm! @caleb531 as you said, the self type was introduced for exactly this situation

PR #142, opened 2023-04-12
More type hints and slots fix

eliotwrobson:
@caleb531 see title. Fixed an issue with slots and added types + simplified a function in the MNTM class.


caleb531 (approved):
Everything here looks great to me!

PR #143, opened 2023-04-14
Refactor *Stack / *Configuration classes to be dataclass-based

caleb531:
@eliotwrobson Alright, here's my attempt to refactor the PDAStack and `*Configuration` classes to use dataclasses instead of namedtuple inheritance. 

### The return of tuple conversion

I decided to keep the casting to a tuple on instantiation because I felt it was more consistent with the type-casting we are performing currently for the automaton subclasses. Plus, it is just easier to work with because I don't have to remember to explicitly pass a tuple to the constructor

### Inheriting from `Tuple`?

As I was looking at the `PDAStack` class, I was noticing how the `PDAStack` is basically a glorified wrapper for the `stack` tuple underneath. Semantically, they represent the same thing, with no difference in data but only difference in behavior. Therefore, I was thinking: maybe we could just inherit directly from `Tuple[str, ...]`, which would eliminate the redeclaration of `__len__`, give us indexing for free, and just feel less redundant.

For example, if you wanted to create a PDAStack and access its elements, you currently have to write something like:

```python
pda_stack = PDAStack(('a', 'b', 'c'))
print(pda_stack.stack[0])  # the property access here makes the semantics feel a little redundant
```

Versus if we inherit directly from `Tuple`:

```python
stack = PDAStack(('a', 'b', 'c'))
print(stack[0])
```

Now, I suppose we could solve the semantics issue somewhat by renaming the `stack` attribute to something like `elements` (hence, the first example can become something like `stack.elements[0]`. But I would argue this is still redundant, since a `PDAStack` stores no additional data members than an actual `tuple`.

Compare this to a `PDAConfiguration` instance, which stores multiple data members and therefore cannot be simplified any further (i.e. its existence is more justified), at least not without losing the labelling of fields and such. But it just seems like the only benefit to using a `PDAStack` as opposed to a regular tuple is the additional behavior provided via new methods.

## Regarding the *Configuration classes

I've also `PDAConfiguration`, `TMConfiguration`, and `MTMConfiguration` classes to be proper dataclasses instead of named tuples. The side effect of doing this is that we can no longer access the members of the object using integer-based indexing (which apparently I was relying on in my tests). I actually like this, since I think attribute names are less ambiguous anyway, so I updated my tests accordingly (e.g. `config.remaining_input` instead of `config[1]`).

## What about the TMTape class?

Totally forgot about this. Will get to this later or in a separate PR.

### DIFF
--- a/automata/pda/stack.py
+++ b/automata/pda/stack.py
@@ -2,19 +2,18 @@
 """Classes and methods for working with PDA stacks."""
 from __future__ import annotations
 
-import collections
+from dataclasses import dataclass
 from typing import Iterator, Sequence, Tuple
 
 
-class PDAStack(collections.namedtuple("PDAStack", ["stack"])):
+@dataclass(frozen=True, slots=True)
+class PDAStack:
     """A PDA stack."""
 
     stack: Tuple[str, ...]
 
-    # TODO can we get rid of this? Kinda weird inheritance here
-    def __new__(cls, elements: Sequence[str]):
-        """Create the new PDA stack."""
-        return super(PDAStack, cls).__new__(cls, tuple(elements))
+    def __init__(self, elements: Sequence[str]):
+        object.__setattr__(self, "stack", tuple(elements))
 
     def top(self) -> str:
         """Return the symbol at the top of the stack."""
@@ -50,4 +49,4 @@ class PDAStack(collections.namedtuple("PDAStack", ["stack"])):
 
     def __repr__(self) -> str:
         """Return a string representation of the stack."""
-        return "{}{}".format(self.__class__.__name__, self.stack)
+        return "{}({})".format(self.__class__.__name__, self.stack)
--- a/automata/pda/stack.py
+++ b/automata/pda/stack.py
@@ -6,9 +6,11 @@ from dataclasses import dataclass
 from typing import Iterator, Sequence, Tuple
 
 
-@dataclass(frozen=True, slots=True)
+@dataclass(frozen=True)
 class PDAStack:
     """A PDA stack."""
+
+    __slots__ = ("stack",)
 
     stack: Tuple[str, ...]
 
--- a/automata/pda/stack.py
+++ b/automata/pda/stack.py
@@ -14,8 +14,8 @@ class PDAStack:
 
     stack: Tuple[str, ...]
 
-    def __init__(self, elements: Sequence[str]):
-        object.__setattr__(self, "stack", tuple(elements))
+    def __init__(self, stack: Sequence[str]):
+        object.__setattr__(self, "stack", tuple(stack))
 
     def top(self) -> str:
         """Return the symbol at the top of the stack."""
--- a/automata/pda/stack.py
+++ b/automata/pda/stack.py
@@ -14,7 +14,7 @@ class PDAStack:
 
     stack: Tuple[str, ...]
 
-    def __init__(self, stack: Sequence[str]):
+    def __init__(self, stack: Sequence[str]) -> None:
         object.__setattr__(self, "stack", tuple(stack))
 
     def top(self) -> str:
--- a/automata/tm/tape.py
+++ b/automata/tm/tape.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Classes and methods for working with Turing machine tapes."""
 
-import collections
+from dataclasses import dataclass
 from typing import Iterator, Sequence, Tuple
 
 from typing_extensions import Self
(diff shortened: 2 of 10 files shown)


automata/pda/stack.py
```diff
@@ -2,19 +2,20 @@
...
 
+    __slots__ = ("stack",)
+
     stack: Tuple[str, ...]
 
-    # TODO can we get rid of this? Kinda weird inheritance here
-    def __new__(cls, elements: Sequence[str]):
-        """Create the new PDA stack."""
-        return super(PDAStack, cls).__new__(cls, tuple(elements))
+    def __init__(self, elements: Sequence[str]):
```
eliotwrobson, line 17: I think annotating this as a sequence and conversion to a tuple makes the most sense (better than what was done before).

tests/test_dtm.py
```diff
@@ -265,24 +265,24 @@ def test_validate_blank_symbol(self):
     def test_read_input_accepted(self):
         """Should return correct state if acceptable TM input is given."""
         final_config = self.dtm1.read_input("00001111")
-        self.assertEqual(final_config[0], "q4")
-        self.assertEqual(str(final_config[1]), "TMTape('xxxxyyyy..', 9)")
+        self.assertEqual(final_config.state, "q4")
```
eliotwrobson, line 268: As you stated in your comment, using the named accessors here is probably best.

eliotwrobson (commented):
Changes here look good so far. I'm not sure I like inheriting directly from tuple in some places (feels like trying to maybe be too clever). Didn't explicitly approve because I think it makes sense to bundle in the TM tape changes you mentioned in the comment.

automata/pda/stack.py
caleb531: @eliotwrobson Me too. I guess it's a benefit of no longer inheriting from `namedtuple` (although my thought was that if we inherit from `Tuple`, the conversion happens automatically and we no longer need any explicit `__init__` or `__new__` definition whatsoever).

eliotwrobson: I think that might make sense, because the internal functions all treat this class as immutable.

caleb531: @eliotwrobson So ironically, I actually just thought of an argument NOT to inherit from tuple:

If we ever decide in the future that `PDAStack` should have any additional data members, we would need to resort back to adding a `__new__` definition (with `super()`) because `tuple()` only accepts a single argument natively. Maybe that's not an issue now while `PDAStack` still has a single data member `stack`, but trying to consider a solution that is future-proof.

I was also trying to think about consistency with similar sequence-type structures like `TMTape`, which accepts a `tape` parameter as input. So if we make `PDAStack` inherit from `Tuple`, would it suggest we should make `TMTape` inherit from `tuple` as well? But `TMTape` requires multiple arguments, so we'd have to use `__new__` in that case.

Maybe I'm just overthinking this in some futile pursuit of API perfection. I suppose it really shouldn't matter much, since `PDAStack` and friends aren't publicly-documented classes (beyond being part of return values, anyway). But gosh, I just feel so indecisive about this now lol. I feel like I could go either way.

eliotwrobson: @calebevans Ahh I see what you mean. I lean towards consistency with the TMTape and not having to define the __new__ function, since I'm generally a fan of doing as few things under the hood as possible as to make code more readable and easier to debug.

calebevans: @eliotwrobson wrong caleb

automata/tm/tape.py
```diff
@@ -1,33 +1,35 @@
...
 
+    __slots__ = ("tape", "blank_symbol", "current_position")
+
     tape: Tuple[str]
     blank_symbol: str
     current_position: int
 
-    def __new__(
-        cls, tape: Sequence[str], *, blank_symbol: str, current_position: int = 0
+    def __init__(
```
eliotwrobson, line 22: Missing type annotations on some functions in this PR. Without this, mypy won't check the internals.

eliotwrobson (commented):
Everything looks good now

automata/pda/stack.py
caleb531: @eliotwrobson This is why it's so nice to have you on the project now. 😌

You have tipped the scale for me, and I am in agreement to avoid inheriting from native types here (and perhaps as a rule-of-thumb going forward, as well).

And I guess the best part of this that by just leaving the object structure as-is, we are not changing the public (or private) API in this respect. One less thing to document lol

automata/tm/tape.py
caleb531: @eliotwrobson Thanks for the catch. I pushed up a08100c42d32dfe9752270d49fdb9d2a8df76a46 where (I think) I got them all.

eliotwrobson: Sg! Once we get everything in the library, it might be good to turn on the no untyped defs flag in mypy.

caleb531: @eliotwrobson I agree! But we'll first need to add types to tests.

PR #145, opened 2023-04-18
Start adding type hints to test cases

eliotwrobson:
@caleb531 See title. A number of small changes accompanied this, to make the adoption of type hints simpler.

- `GNFA` no longer inherits from `NFA`. There was very little actual logic sharing between these classes. This eliminated the need for some redundant tests.
- Some minor changes to the internals of the regex lexer (some methods were just not needed).
- Removed `__reversed__` method for `NFA` class, since this method has different intended usage (should be used on sequences).

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -821,7 +821,7 @@ class DFA(fa.FA):
 
     def successor(
         self,
-        input_str: str,
+        input_str: Optional[str],
         *,
         strict: bool = True,
         key: Optional[Callable[[Any], Any]] = None,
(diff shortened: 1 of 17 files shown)


eliotwrobson:
@caleb531 just added documentation of the changes associated with this PR. If you don't have any objections to the changes at a high level, I can go ahead and merge. A detailed review isn't super necessary (most of the line changes here are just adding simple type hints and making tests compliant).

caleb531 (approved):
Gave this a quick skim—looks fine to me. I trust your work, so I'm not too concerned with scrutinizing every single line of code, especially if all the CI jobs are passing (which they are).

PR #146, opened 2023-04-23
First examples

### LINKED ISSUE #103: Adding directory for usage examples?
So the API documentation reorg (#98 and #99) is complete—great! But as I look through the API documentation, I realize that there is a lot of text to sift through between each block of code.

@Tagl @eliotwrobson What do you think about adding an `examples/` directory with Python files that demonstrate the code needed to perform the most common DFA/etc. operations. Since the API is so large at this point, we would probably need to be selective about what use cases we plan to demonstrate.

One option is that we could make each example built around a real-world use case. For example:

```python 
# examples/reading_input_from_cli.py
from automata.fa.dfa import DFA

# DFA which matches all binary strings ending in an odd number of '1's
my_dfa = DFA(
    states={'q0', 'q1', 'q2'},
    input_symbols={'0', '1'},
    transitions={
        'q0': {'0': 'q0', '1': 'q1'},
        'q1': {'0': 'q0', '1': 'q2'},
        'q2': {'0': 'q2', '1': 'q1'}
    },
    initial_state='q0',
    final_states={'q1'}
)

try:
    while True:
        if my_dfa.accepts_input(input('Please enter your input: ')):
            print('Accepted')
        else:
            print('Rejected')
except KeyboardInterrupt:
    print('')
```

Another option is that we could make each example a sort of rapid-fire walkthrough of the different things you can do. For example:

```python
# examples/dfa_reading_input.py
from automata.fa.dfa import DFA

# DFA which matches all binary strings ending in an odd number of '1's
my_dfa = DFA(
    states={'q0', 'q1', 'q2'},
    input_symbols={'0', '1'},
    transitions={
        'q0': {'0': 'q0', '1': 'q1'},
        'q1': {'0': 'q0', '1': 'q2'},
        'q2': {'0': 'q2', '1': 'q1'}
    },
    initial_state='q0',
    final_states={'q1'}
)

my_dfa.read_input('001') # returns a final state (e.g. 'q1')

my_dfa.read_input('001!') # raises RejectionException

# loop over each state of the DFA as you read input
for state in my_dfa.read_input_stepwise('00111'):
    print(state)

# etc.
```

We may also need to be careful of adding too many examples, such that it becomes an additional maintenance burden on top of the existing API documentation. But it might be helpful for beginners. I'm a bit ambivalent—what do you think?

eliotwrobson:
Added some early examples for `FA` classes. Resolves #103 for these classes.

This is just meant to be a starting point. Happy to add any examples people might want to donate, and leaning towards merging this soon and adding onto it before the next release.

### DIFF
--- a/docs/fa/class-dfa.md
+++ b/docs/fa/class-dfa.md
@@ -455,5 +455,4 @@ dfa.show_diagram(path='./dfa1.png')
 
 [Next: NFA Class](class-nfa.md)  
 [FA Class](class-fa.md)  
-[Examples](examples.md)  
 [Table of Contents](../README.md)
--- a/docs/fa/class-dfa.md
+++ b/docs/fa/class-dfa.md
@@ -455,4 +455,5 @@ dfa.show_diagram(path='./dfa1.png')
 
 [Next: NFA Class](class-nfa.md)  
 [FA Class](class-fa.md)  
+[Examples](examples.md)  
 [Table of Contents](../README.md)
--- a/docs/fa/examples.md
+++ b/docs/fa/examples.md
@@ -1,0 +1,11 @@
+# Examples
+
+[FA Class](class-fa.md)  
+[NFA Class](class-nfa.md)  
+[DFA Class](class-dfa.md)  
+[Table of Contents](../README.md)
+
+
+------
+
+[Table of Contents](../README.md)
--- a/docs/fa/examples.md
+++ b/docs/fa/examples.md
@@ -1,15 +1,15 @@
-# Examples
+# FA Examples
 
 [FA Class](class-fa.md)  
-[NFA Class](class-nfa.md)  
-[DFA Class](class-dfa.md)  
 [Table of Contents](../README.md)
 
-On this page, we give some short examples with discussion for the finite automata classes and methods.
+On this page, we give some short examples with discussion for the finite
+automata classes and methods.
 
 ## Reading basic input
 
-The following code snippet creates a `DFA` and prints whether it accepts or rejects user input.
+The following code snippet creates a `DFA` and prints whether it accepts or
+rejects user input.
 
 ```python
 from automata.fa.dfa import DFA
@@ -39,10 +39,12 @@ except KeyboardInterrupt:
 
 ## Subset for NFAs
 
-The `NFA` does not have a built-in method for checking whether it is a subset of another `NFA`. However, this can be done using existing methods.
+The `NFA` does not have a built-in method for checking whether it is a subset
+of another `NFA`. However, this can be done using existing methods.
 
 ```python
-# In the following, we have nfa1 and nfa2 and want to determine whether nfa1 is a subset of nfa2.
+# In the following, we have nfa1 and nfa2 and want to determine whether
+# nfa1 is a subset of nfa2.
 
 # If taking the union of nfa2 with nfa1 is equal to nfa2 again,
 # nfa1 didn't accept any strings that nfa2 did not, so it is a subset.
@@ -55,7 +57,13 @@ else:
 
 ## Edit distance automaton
 
-The following example is inspired by [this blog post](http://blog.notdot.net/2010/07/Damn-Cool-Algorithms-Levenshtein-Automata). Essentially, given a set of strings, we want to determine whether 
+The following example is inspired by [this blog post][levelshtein-article].
+Essentially, we want to determine which strings in a given set are within
+the target edit distance to a reference string.
+
+[levelshtein-article]: http://blog.notdot.net/2010/07/Damn-Cool-Algorithms-Levenshtein-Automata
+
+
 
 ```python
 from automata.fa.dfa import DFA
@@ -89,7 +97,10 @@ words_within_edit_distance_dfa = DFA.from_nfa(
 found_words_dfa = target_words_dfa & words_within_edit_distance_dfa
 found_words = list(found_words_dfa)
 
-print(f"All words within edit distance {edit_distance} of '{reference_string}': {found_words}")
+print(
+    f"All words within edit distance {edit_distance} of "
+    f"'{reference_string}': {found_words}"
+)
 ```
 
 ------
--- a/docs/fa/examples.md
+++ b/docs/fa/examples.md
@@ -5,6 +5,92 @@
 [DFA Class](class-dfa.md)  
 [Table of Contents](../README.md)
 
+On this page, we give some short examples with discussion for the finite automata classes and methods.
+
+## Reading basic input
+
+The following code snippet creates a `DFA` and prints whether it accepts or rejects user input.
+
+```python
+from automata.fa.dfa import DFA
+
+# DFA which matches all binary strings ending in an odd number of '1's
+my_dfa = DFA(
+    states={'q0', 'q1', 'q2'},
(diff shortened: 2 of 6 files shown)


caleb531:
@eliotwrobson Just realized I totally missed this. Let me review it now!

docs/fa/class-dfa.md
```diff
@@ -455,4 +455,5 @@ dfa.show_diagram(path='./dfa1.png')
 
 [Next: NFA Class](class-nfa.md)  
 [FA Class](class-fa.md)  
+[Examples](examples.md)
```
caleb531, line 458: @eliotwrobson I am hesitant to add Examples to the bottom navigation of each docs page. I see the bottom navigation for one of two things:

1. Next section
2. Page hierarchy

The Examples section doesn't fall into either of those two categories (it's generally a non-adjacent sibling page). Plus, it might imply that the examples are all DFA-specific, whereas it actually encompasses everything for FA.

That's my view anyway. But hey, I'm open to hearing your thoughts on this!

eliotwrobson: Ah, I think that's perfectly sound reasoning. TBH I didn't really have an opinion, so I kinda just threw the links to the examples at the bottom navigation. Will remove a little later.

docs/fa/examples.md
```diff
@@ -0,0 +1,97 @@
+# Examples
+
+[FA Class](class-fa.md)  
+[NFA Class](class-nfa.md)
```
caleb531, line 4: @eliotwrobson Apart from the fact that DFA should be listed before NFA to match the FA subclasses list, I'm not sure why we need these here. If we're gonna have them, we might as well have GNFA too. But per my reasoning in my other comment, I think of the top/bottom navigation as more for the purpose of hierarchy and 'Read Next'.

docs/fa/examples.md
```diff
@@ -0,0 +1,97 @@
...
+except KeyboardInterrupt:
+    print('')
+```
+
+## Subset for NFAs
+
+The `NFA` does not have a built-in method for checking whether it is a subset of another `NFA`. However, this can be done using existing methods.
+
+```python
+# In the following, we have nfa1 and nfa2 and want to determine whether nfa1 is a subset of nfa2.
```
caleb531, line 45: @eliotwrobson I hate to be nitpicky, but are you able to reflow this comment onto two lines? I like to have all my lines fit within 80 characters, particular in the documentation.

docs/fa/examples.md
eliotwrobson: sounds reasonable, I'll leave the table of contents link but remove the others.

docs/fa/examples.md
```diff
@@ -0,0 +1,97 @@
...
+if (nfa1 | nfa2) == nfa2:
+    print('nfa1 is a subset of nfa2.')
+else:
+    print('nfa1 is not a subset of nfa2.')
+
+```
+
+## Edit distance automaton
+
+The following example is inspired by [this blog post](http://blog.notdot.net/2010/07/Damn-Cool-Algorithms-Levenshtein-Automata). Essentially, given a set of strings, we want to determine whether
```
caleb531, line 58: @eliotwrobson When the URL is especially long, I tend to prefer reference-style links, per https://daringfireball.net/projects/markdown/syntax#link:

```markdown
The following example is inspired by [this blog post][levelshtein-article].
Essentially, given a set of strings, we want to determine whether 

[levelshtein-article]: http://blog.notdot.net/2010/07/Damn-Cool-Algorithms-Levenshtein-Automata)
```

Also, the last sentence is incomplete ("We want to determine whether ...").

docs/fa/examples.md
```diff
@@ -0,0 +1,97 @@
...
+            print('Accepted')
+        else:
+            print('Rejected')
+except KeyboardInterrupt:
+    print('')
+```
+
+## Subset for NFAs
+
+The `NFA` does not have a built-in method for checking whether it is a subset of another `NFA`. However, this can be done using existing methods.
```
caleb531, line 42: @eliotwrobson This would obviously be part of a separate PR, but do you think it would be worth adding `issubset`, `__le__`, `issuperset`, etc. to the `NFA` class to create parity with `DFA`?

docs/fa/examples.md
caleb531: @eliotwrobson Sorry, should have been more specific: I would like to keep both FA and Table of Contents links here, since they are technically part of the hierarchy (I see the FA class page in this context as being the Table of Contents page for this `docs/fa` subdirectory).

In other words, I only want to remove the DFA and NFA links here.

docs/fa/examples.md
```diff
@@ -0,0 +1,97 @@
+# Examples
```
caleb531, line 1: @eliotwrobson One more thought: I think it would be good to rename this to `FA Examples` to distinguish between any additional examples pages we create in the future (e.g. for PDA or TM).

docs/fa/examples.md
eliotwrobson: No worries! This is a pretty easy thing to do.

docs/fa/examples.md
eliotwrobson: Done here: https://github.com/caleb531/automata/pull/146/commits/2ce6816e437d8ca17b4e095ab4922e18eb4421f7

docs/fa/examples.md
eliotwrobson: Addressed here: https://github.com/caleb531/automata/pull/146/commits/2ce6816e437d8ca17b4e095ab4922e18eb4421f7

docs/fa/examples.md
eliotwrobson: Done here https://github.com/caleb531/automata/pull/146/commits/2ce6816e437d8ca17b4e095ab4922e18eb4421f7

docs/fa/examples.md
eliotwrobson: Done here https://github.com/caleb531/automata/pull/146/commits/2ce6816e437d8ca17b4e095ab4922e18eb4421f7

docs/fa/class-dfa.md
eliotwrobson: Done here https://github.com/caleb531/automata/pull/146/commits/2ce6816e437d8ca17b4e095ab4922e18eb4421f7

docs/fa/examples.md
eliotwrobson: So in general yes, but I would like to use an algorithm other than the code here. Basically, I think that every built-in method to these classes should be optimized beyond calling other methods that construct other automaton under the hood. For the subset algorithm, there are faster algorithms than the one used here (don't involve constructing a separate NFA), see http://www.lsv.fr/~doyen/papers/Antichains_Algorithms_Finite_Automata.pdf for an algorithm that's supposed to be faster.

However, I think some of these algorithms are tricky to implement and would take some care by someone that knows what they're doing following a paper.

caleb531: @eliotwrobson All good points. Let's not worry about it then, at least for now (certainly not before the eventual v8 release).

eliotwrobson: @caleb531 Sounds good! I'll go ahead and resolve this discussion, and since this is the last item, I'll merge in just a sec.

PR #147, opened 2023-05-08
Proper support for allow_partial

EduardoGoulart1:
@eliotwrobson @caleb531 Please do not review it because the code is not ready yet. This is just a draft to book progress on the allow_partial support.

If you go with the definition that DFAs are sets of words, then all operations are well-defined and relatively easy to implement. Most of the times the only required change is to replace loops like:

```Python
for symbol in input_symbols:
    tgt_a = transitions_a[symbol]
```
By the code:
```Python
for symbol in input_symbols:
    tgt_a = transitions_a.get(tgt_a, None)
```

I made the code such that it adds no overhead for complete DFAs compared to the current implementation. For partial DFAs, if the DFAs are sparse, then you will get a large performance boost, while if they are dense you will pay some penalty (but most of the times negligible).

I feel like we still need to discuss a few things:

1. ~~Reserve something (probably None) to represent the trap state. Basically, we add the invariant that if None is part of the set of states, then it *must* be a trap state. This is already implicitly assumed in `_get_next_current_state` and greatly simplifies the implementation logic~~. Because network graph does not allow for states labeled None, we replaced it with automatically generated integers
2. We could think of making this allow_partial thing invisible to the users. Instead of passing allow_partial as a parameter, we compute internally the flag "is_partial". I go by the principle that we should not restrict our API unless it would lead to misuse.
3. Now that support for partial DFAs is there, it is straightforward to extend operations to support DFAs with different alphabets

**Missing:**
- [ ] Adjust/add tests
- [ ] Add benchmarks

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -307,7 +307,7 @@ class DFA(fa.FA):
     ) -> Self:
 
         added_trap_state = False
-        new_transitions: Dict[DFAStateT, DFAPathT] = {}
+        new_transitions: Dict[DFAStateT, Dict[str, DFAStateT]] = {}
 
         for state, state_path in transitions.items():
             new_state_path = new_transitions.setdefault(state, state_path)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1241,7 +1241,7 @@ class DFA(fa.FA):
         prefix: str,
         *,
         contains: bool = True,
-        as_partial: bool = False,
+        as_partial: bool = True,
     ) -> Self:
         """
         Directly computes the minimal DFA recognizing strings with the
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1834,7 +1834,7 @@ class DFA(fa.FA):
             current_state_name = get_name(current_states)
 
             # Add NFA states to DFA as it is constructed from NFA.
-            dfa_transitions[current_state_name] = {}
+            path_dict = dfa_transitions.setdefault(current_state_name, {})
             if not current_states.isdisjoint(target_nfa.final_states):
                 dfa_final_states.add(current_state_name)
 
@@ -1843,10 +1843,9 @@ class DFA(fa.FA):
                 input_symbol,
                 next_current_states,
             ) in target_nfa._iterate_through_symbol_path_pairs(current_states):
+                
                 next_current_states_name = get_name(next_current_states)
-                dfa_transitions[current_state_name][
-                    input_symbol
-                ] = next_current_states_name
+                path_dict[input_symbol] = next_current_states_name
 
                 # Only enqueue a state if it has not been seen yet.
                 if next_current_states_name not in dfa_states:
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -2,6 +2,7 @@
 """Classes and methods for working with deterministic finite automata."""
 from __future__ import annotations
 
+import array
 from collections import defaultdict, deque
 from itertools import chain, count
 from random import Random
@@ -1397,7 +1398,9 @@ class DFA(fa.FA):
         If contains is set to False then the complement is constructed instead.
         If must_be_suffix is set to True, then the substring must be a suffix instead.
         """
-        transitions: DFATransitionsT = {i: {} for i in range(len(substring))}
+        transitions: Dict[DFAStateT, Dict[str, DFAStateT]] = {
+            i: {} for i in range(len(substring))
+        }
         transitions[len(substring)] = {
             symbol: len(substring) for symbol in input_symbols
         }
@@ -1405,7 +1408,7 @@ class DFA(fa.FA):
         # Computing failure function for partial matches as is done in the
         # Knuth-Morris-Pratt string algorithm so we can quickly compute the
         # next state from another state
-        kmp_table = [-1 for _ in substring]
+        kmp_table = array.array("i", (-1 for _ in substring))
         candidate = 0
         for i, char in enumerate(substring):
             if i == 0:
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1843,22 +1843,15 @@ class DFA(fa.FA):
                 input_symbol,
                 next_current_states,
             ) in target_nfa._iterate_through_symbol_path_pairs(current_states):
-                #print(input_symbol, next_current_states)
-                # next_current_states = target_nfa._get_next_current_states(
-                #    current_states, input_symbol
-                # )
+                next_current_states_name = get_name(next_current_states)
+                dfa_transitions[current_state_name][
+                    input_symbol
+                ] = next_current_states_name
 
-                # Can ignore trivial trap state (going to the empty set)
-                if next_current_states:
-                    next_current_states_name = get_name(next_current_states)
-                    dfa_transitions[current_state_name][
-                        input_symbol
-                    ] = next_current_states_name
-
-                    # Only enqueue a state if it has not been seen yet.
-                    if next_current_states_name not in dfa_states:
-                        dfa_states.add(next_current_states_name)
-                        state_queue.append(next_current_states)
+                # Only enqueue a state if it has not been seen yet.
+                if next_current_states_name not in dfa_states:
+                    dfa_states.add(next_current_states_name)
+                    state_queue.append(next_current_states)
 
         # if minify:
         #    return cls._minify(
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1737,11 +1737,6 @@ class DFA(fa.FA):
         def subset_function(current_states: FrozenSet[DFAStateT]) -> bool:
             return not current_states.isdisjoint(target_nfa.final_states)
 
-        def expand_state_fn(
-            current_states: FrozenSet[DFAStateT],
-        ) -> ExpandStateReturnType:
-            yield from target_nfa._iterate_through_symbol_path_pairs(current_states)
-
         initial_state = frozenset(
             target_nfa._get_lambda_closures()[target_nfa.initial_state]
         )
@@ -1749,7 +1744,7 @@ class DFA(fa.FA):
         result = cls._expand_dfa(
             subset_function,
             initial_state,
-            expand_state_fn,
+            target_nfa._iterate_through_symbol_path_pairs,
             target_nfa.input_symbols,
             retain_names=retain_names,
             minify=minify,
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1560,7 +1560,7 @@ class DFA(fa.FA):
         cls: Type[Self],
         input_symbols: AbstractSet[str],
         language: AbstractSet[str],
-        as_partial_dfa: bool = True,
+        as_partial: bool = True,
     ) -> Self:
         """
         Directly computes the minimal DFA corresponding to a finite language.
@@ -1651,14 +1651,14 @@ class DFA(fa.FA):
 
         compress(prev_word, "")
 
-        if as_partial_dfa:
+        if as_partial:
             return cls(
                 states=frozenset(transitions.keys()),
                 input_symbols=input_symbols,
                 transitions=transitions,
                 initial_state="",
                 final_states=final_states,
-                allow_partial=as_partial_dfa,
+                allow_partial=as_partial,
             )
 
         return cls._to_complete(
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -783,6 +783,7 @@ class DFA(fa.FA):
         """
 
         if retain_names:
+
             def get_name_original(state: DFAStateT) -> DFAStateT:
                 return state
 
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -238,7 +238,7 @@ class DFA(fa.FA):
         return self.cardinality()
 
     # Supports partial (test)
-    def as_partial(self) -> Self:
+    def to_partial(self) -> Self:
         """
         Turns a DFA (complete or not) into a partial DFA.
         Removes dead states and trap states (except the initial state)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -858,6 +858,7 @@ class DFA(fa.FA):
         """
 
         if retain_names:
+
             def get_name_original(state: DFAStateT) -> DFAStateT:
                 return state
 
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -273,9 +273,6 @@ class DFA(fa.FA):
             allow_partial=True,
         )
 
-    def get_trap_state_id(self) -> int:
-        return next(x for x in count(-1, -1) if x not in self.states)
-
     # Supports partial (test)
     def to_complete(self) -> Self:
         """
@@ -285,22 +282,42 @@ class DFA(fa.FA):
         if not self.allow_partial:
             return self.copy()
 
-        trap_state = self.get_trap_state_id()
-        default_to_trap: DFAPathT = {
-            symbol: trap_state for symbol in self.input_symbols
-        }
-        transitions: Dict[DFAStateT, DFAPathT] = {
-            state: {**default_to_trap, **lookup}
-            for state, lookup in self.transitions.items()
-        }
-        transitions[trap_state] = default_to_trap
+        return self._to_complete(states=self.states)
 
-        return self.__class__(
-            states=frozenset(transitions.keys()),
-            input_symbols=self.input_symbols,
-            transitions=transitions,
-            initial_state=self.initial_state,
-            final_states=self.final_states,
+    @classmethod
+    def _to_complete(
+        cls: Type[Self],
+        *,
+        states: AbstractSet[DFAStateT],
+        input_symbols: AbstractSet[str],
+        transitions: DFATransitionsT,
+        initial_state: DFAStateT,
+        final_states: AbstractSet[DFAStateT],
+        trap_state: Optional[DFAStateT],
+    ) -> Self:
+        if trap_state is None:
+            trap_state = next(x for x in count(-1, -1) if x not in states)
+
+        added_trap_state = False
+        new_transitions: Dict[DFAStateT, DFAPathT] = {}
+
+        for state, state_path in transitions.items():
+            new_state_path = new_transitions.setdefault(state, state_path)
+
+            for symbol in input_symbols:
+                if symbol not in new_state_path:
+                    new_state_path[symbol] = trap_state
+                    added_trap_state = True
+
+        if added_trap_state:
+            transitions[trap_state] = {symbol: trap_state for symbol in input_symbols}
+
+        return cls(
+            states=frozenset(new_transitions.keys()),
+            input_symbols=input_symbols,
+            transitions=new_transitions,
+            initial_state=initial_state,
+            final_states=final_states,
             allow_partial=False,
         )
 
@@ -1839,13 +1856,13 @@ class DFA(fa.FA):
                 retain_names=retain_names,
             )
 
-        return cls(
+        return cls._to_complete(
             states=dfa_states,
             input_symbols=dfa_symbols,
             transitions=dfa_transitions,
             initial_state=dfa_initial_state,
             final_states=dfa_final_states,
-            allow_partial=False,
+            trap_state=frozenset(),
         )
 
     def iter_transitions(
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1386,10 +1386,9 @@ class DFA(fa.FA):
 
         if not as_partial:
             err_state = -1
-            for i, _ in enumerate(prefix):
-                transitions_i = transitions[i]
+            for state_path in transitions.values():
                 for symbol in input_symbols:
-                    transitions_i.setdefault(symbol, err_state)
+                    state_path.setdefault(symbol, err_state)
             transitions[err_state] = {symbol: err_state for symbol in input_symbols}
 
         states = frozenset(transitions.keys())
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1179,14 +1179,11 @@ class DFA(fa.FA):
         """
         Populate word cache up to length k
         """
-        if not self.allow_partial:
-            sorted_symbols = sorted(self.input_symbols)
-            sorted_transition_symbols = {state: sorted_symbols for state in self.states}
-        else:
-            sorted_transition_symbols = {
-                state: sorted(lookup.keys())
-                for state, lookup in self.transitions.items()
-            }
+
+        # Weird construction to account for partial DFAs.
+        sorted_transition_symbols = {
+            state: sorted(lookup.keys()) for state, lookup in self.transitions.items()
+        }
 
         while len(self._word_cache) <= k:
             i = len(self._word_cache)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -595,31 +595,11 @@ class DFA(fa.FA):
             q_a, q_b = state_pair
             return q_a in self.final_states or q_b in other.final_states
 
-        initial_state, expand_state_fn = self.__class__._cross_product(self, other)
-
-        if self.allow_partial or other.allow_partial:
-            trap_a = self.get_trap_state_id()
-            trap_b = other.get_trap_state_id()
-
-            def partial_union_expand_state_fn(
-                state: DFAStateT,
-            ) -> ExpandStateReturnType:
-                q_a, q_b = state
-                q_a_transitions = self.transitions.get(q_a, {})
-                q_b_transitions = self.transitions.get(q_b, {})
-
-                for chr, tgt_a in q_a_transitions.items():
-                    tgt_b = q_b_transitions.get(chr, trap_b)
-                    yield chr, (tgt_a, tgt_b)
-                for chr, tgt_b in q_b_transitions.items():
-                    if chr not in q_a_transitions:
-                        yield chr, (trap_a, tgt_b)
-
-            expand_state_fn = partial_union_expand_state_fn
+        expand_state_fn = self.__class__._cross_product(self, other, True, True)
 
         return self.__class__._expand_dfa(
             union_function,
-            initial_state,
+            (self.initial_state, other.initial_state),
             expand_state_fn,
             self.input_symbols,
             retain_names=retain_names,
@@ -640,27 +620,11 @@ class DFA(fa.FA):
             q_a, q_b = state_pair
             return q_a in self.final_states and q_b in other.final_states
 
-        initial_state, expand_state_fn = self.__class__._cross_product(self, other)
-
-        # We only expand characters in both transition functions
-        if self.allow_partial or other.allow_partial:
-
-            def partial_intersection_expand_state_fn(
-                state: DFAStateT,
-            ) -> ExpandStateReturnType:
-                q_a, q_b = state
-                transitions_a = self.transitions[q_a]
-                transitions_b = other.transitions[q_b]
-
-                for chr, tgt_a in transitions_a.items():
-                    if chr in transitions_b:
-                        yield chr, (tgt_a, transitions_b[chr])
-
-            expand_state_fn = partial_intersection_expand_state_fn
+        expand_state_fn = self.__class__._cross_product(self, other, False, False)
 
         return self.__class__._expand_dfa(
             intersection_function,
-            initial_state,
+            (self.initial_state, other.initial_state),
             expand_state_fn,
             self.input_symbols,
             retain_names=retain_names,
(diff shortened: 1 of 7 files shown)


automata/fa/dfa.py
```diff
@@ -211,10 +223,75 @@ def __iter__(self) -> Iterator[str]:
...
+    def as_partial(self) -> DFA:
+        """
+        Turns a DFA (complete or not) into a partial DFA.
+        Removes dead states and trap states (except the initial state)
+            and all edges leading to them.
+        """
+        if self.allow_partial:
+            return self.copy()
+
+        graph = self._get_digraph()
```
eliotwrobson, line 241: Now that I'm thinking of it, the digraph associated with a DFA should probably be made into a cached property.

EduardoGoulart1: It is funny because I was going to propose quite the opposite. Getting rid of any cached property

EduardoGoulart1: We don't need to discuss it here, but my problem is specifically with the word cache. Caching all words of length k is a (silent) memory bomb. In my opinion, this should be implemented as an iterator and if the users want fast access to this set, they should compute and store it themselves.

Caching digraph and word lengths seem to offer an acceptable memory overhead (I would still vote for its removal otherwise we need to worry about keeping it consistent)

caleb531: I do share @EduardoGoulart1's concern about bloating memory usage, but am not sure how to balance it if the alternative comes with a considerable performance impact to many methods. Maybe we make this a global configuration option (i.e. in [`automata.base.config`](https://github.com/caleb531/automata/blob/main/automata/base/config.py)) and leave it up to the user?

I'm a bit undecided on the right path forward, so @eliotwrobson, please let me know your thoughts, and perhaps we can all come to an agreement on the right approach.

eliotwrobson: I think either a global or local config is the way to go. Kinda like the immutability changes, for most situations, the current memory usage isn't a huge deal, and it makes common operations run quickly (I think that that the accepted words are actually computed using dynamic programming, so memory usage at some level is unavoidable). But I agree that it would be nice to have the option to turn these off if people are running into performance issues. Setting this via global or local config should be fairly straightforward.

I don't think this quite applies yet, but if people are running into more and more performance issues, we could consider introducing Cython as a backend for this project. That will _substantially_ improve things on both of these fronts. However, this would be a significant effort, so maybe something to consider in v9 if this is still a concern.

caleb531: Okay, great! Since we have a couple different scenarios:

1. Cache words
2. Cache diggraph / word lengths
3. Disabled cache

I wonder if we should have some config property like `cache_mode` or `word_cache_mode` (@eliotwrobson, maybe help me with a good name here 😅), with an enum as the set of possible values.

As for the Cython thing, I have zero experience writing C extensions for Python. It does sound like a very interesting idea though, provided we have someone who knows what they're doing (not sure how familiar you yourself are).

eliotwrobson: > Okay, great! Since we have a couple different scenarios:
> 
> 1. Cache words
> 2. Cache diggraph / word lengths
> 3. Disabled cache

@caleb531 maybe call this a `cache_level` enum, where higher levels equate to more caching (so cache everything > just cache the digraph > cache nothing). I apologize somewhat for derailing discussion here a bit, it's probably best to handle this after everything else for v8 has already been merged (since the code changes are going to be short and this is a performance improvement and not a new feature).

> As for the Cython thing, I have zero experience writing C extensions for Python. It does sound like a very interesting idea though, provided we have someone who knows what they're doing (not sure how familiar you yourself are).

I'm not really familiar with this either (although I actually think it's fairly straightforward, the code I've seen just looks like Python with some extra information). I think I mentioned this somewhere else, but most of the current implementations have really hit the limit of performance in terms of what can be done in pure Python. Everything here is efficient enough for my purposes, so I think this can wait if/when someone comes along needing to really push the efficiency and wants to take this on.

@EduardoGoulart1 speaking of which, if you find yourself needing minimization for DFAs over large alphabets, there actually is a specialized algorithm that performs better than the one implemented here. So even for this use case, we don't have to resort to introducing new technologies quite yet 😅 .

EduardoGoulart1: Not sure about how the Cython extensions could look like, but I would prefer to avoid it as much as possible... 
@eliotwrobson which algorithm are you talking about? I only know [this one](https://www.sciencedirect.com/science/article/abs/pii/S0020019011003267)

EduardoGoulart1: Let's bring this discussion somewhere else, I think that for now we can just go along with this PR (I'll try to wrap it up the next days)

EduardoGoulart1:
I'm currently working on finalizing this PR and I value your opinions before proceeding.

When transitioning from partial to complete DFAs, it is necessary to include a trap state. So far I was using None for this purpose, but it doesn't work well with networkx.DiGraph because that library doesn't allow nodes labeled None in the graph. It is not too difficult to work around this problem for the DFA class, but it becomes problematic when calling NFA.to_dfa (or other automata types) as this invariant would need to propagate throughout the code. So I need another approach to generate new trap states.

The only solution I can think of that does not require significant refactoring, is to assign the highest negative integer number that isn't already a state. This solution works well, but sometimes states are strings or frozen sets, which leads to a typing mismatch that isn't directly caused by the user. To some extent, this can be mitigated on our side, but I don't think that we can completely avoid it.

caleb531:
@EduardoGoulart1 Regarding mixed state types, that's a fair question. @eliotwrobson I think we've updated the library to handle, for example, states of mixed types, correct? I believe some existing methods (maybe `DFA.minify`?) produce machines with such mixed state types.

If this is the case, then I'm fine with the highest negative integer number.

eliotwrobson:
@caleb531 yes, the library can handle mixed state types, and there are a couple of places this is used I think (see `from_finite_language`). @EduardoGoulart1 I think that's a fine convention, as there are a few places where we start assigning integer state names starting from 0 (in the regex code I believe).

EduardoGoulart1:
Also, adapting the test is fairly easy, but some tests have hard-coded values for the expected result which for me does not make sense. Here for instance this test is explicitly testing the union operator by constructing the DFA representing the result.

Is there a specific reason for that? I would have expected the test to verify its algebraic properties. For example something like `A \subseteq (A \cup B) && B \subseteq (A \cub B) && (A \cub B) - A - B == \emptyset`

image

eliotwrobson:
@EduardoGoulart1 the reason for these is to verify the state names and types of the output (which when names are retained, is part of the API). These test cases should be kept with their hard-coded output. There actually is a test case testing for algebraic properties separately.

I'm not sure about the best way to test with allow partial for these. Is there a way to use parameterized test cases (like in pytest) with nose? The DFAs in those test cases are not partial anyway, so I don't think they can really be meaningfully adapted to test for the new behavior in this PR.

caleb531:
@eliotwrobson @EduardoGoulart1 It does look like nose2 supports test parameterization! Although since DFAs can be verbose to construct, I wonder if the decorator signature would get pretty large. But please feel free to play around to find something practical and maintainable:

https://docs.nose2.io/en/latest/params.html

EduardoGoulart1:
@caleb531 @eliotwrobson sure my initial idea was to extend the tests to use dfa.as_partial(). The only places where this does not work is for such hard-coded tests. The problem is that we do not expand all the states. But I will try my best with these constraints in mind :D

eliotwrobson:
@EduardoGoulart1 just a heads up that, because of the size of the refactor and some weirdness that was uncovered, I want to wait until #129 is merged before merging this. I fixed the last major blocker there, so hopefully that will happen soon.

caleb531:
@EduardoGoulart1 @eliotwrobson Just merged #129! I'm working on resolving the merge conflicts now.

automata/fa/dfa.py
```diff
@@ -313,16 +399,20 @@ def read_input_stepwise(
...
         """Return a digraph corresponding to this DFA with transition symbols ignored"""
-        return nx.DiGraph(
+        graph = nx.DiGraph(
             [
                 (start_state, end_state)
                 for start_state, transition in self.transitions.items()
                 for end_state in transition.values()
             ]
         )
+        graph.update(nodes=(state for state in self.states))
```
eliotwrobson, line 412:
```suggestion
        graph.update(self.states)
```
Can this just take in any iterable?

eliotwrobson: Updated here https://github.com/caleb531/automata/pull/147/commits/7123f7ec24d8ec0fe1555e1c61aeb213880142d3

eliotwrobson:
@EduardoGoulart1 Some merge conflicts came up when #152 was merged. I've resolved those and gotten the lint check passing in the latest push (with some minor style changes and optimizations). Please pull when you get the chance.

automata/fa/dfa.py
```diff
@@ -1476,22 +1753,25 @@ def compress(word: str, next_word: str) -> None:
...
             input_symbols=input_symbols,
             transitions=transitions,
             initial_state="",
             final_states=final_states,
+            allow_partial=as_partial_dfa,
         )
 
+    # Supports partial
     @classmethod
     def from_nfa(
```
eliotwrobson, line 1776: @EduardoGoulart1 This function should have an `as_partial` option. This is because we always know that the state corresponding to the empty set is a trap state, so if `as_partial` is set, we can always omit and avoid traversing the empty set in the BFS. This should substantially speed up this algorithm when used for converting NFAs that come from regexs (since these NFAs have a lot of transitions to the empty set for unused characters).

EduardoGoulart1: Ok will do... I am of the more radical opinion that there should only be partial DFAs

eliotwrobson: This should be pretty straightforward, as I added some of this functionality in my commits.

Also, I actually think that's not such a radical idea. We can discuss semantics when more things are implemented, but there might be a nice way of making the partial property appear in more places.

caleb531:
@eliotwrobson @EduardoGoulart1 What's outstanding for this ticket before it can be reviewed/merged? I believe this is the last major piece that is necessary for the v8 release.

eliotwrobson:
@caleb531 the changes I made were pretty minor on top of what was here to start. Per previous comments, I think all that's left is to write tests for everything. Not sure how far along @EduardoGoulart1 was with that.

EduardoGoulart1:
Yes, basically the tests... I stopped halfway through them because it's quite a lot to go over :sweat:

eliotwrobson:
> I stopped halfway through them because it's quite a lot to go over

@EduardoGoulart1 could you push what you have (even if partially completed)? This should speed up the review process at least somewhat.

caleb531:
@EduardoGoulart1 Can you please make sure you've pushed up your latest work so @eliotwrobson can finish the tests? I'd love to get this reviewed/merged soon.

automata/fa/dfa.py
```diff
@@ -438,19 +557,24 @@ def _minify(
         new_transitions = {
             name: {
                 letter: back_map[transitions[next(iter(eq))][letter]]
-                for letter in input_symbols
+                for letter, tgt_state in transitions[next(iter(eq))].items()
```
eliotwrobson, line 560: @EduardoGoulart1 Why make this change? Doesn't seem necessary.

eliotwrobson:
@EduardoGoulart1 @caleb531 in the latest push, I streamlined and update the test cases to incorporate partial DFAs, including tests that check for interoperability with partial and non-partial DFAs. This bought the line coverage up to 100% in the DFA file.

There are a couple of minor things left to do. The main thing is double checking over the test suite to ensure we are testing extensively enough for the partial behavior. The other thing is doing a pass over the DFA code to clean things up as much as possible. There were places where I removed a large amount of code, but I may have missed some things.

Even though I'm not quite ready to call this done, I think this is in a mature enough state that we can start reviewing, as it should be mostly code complete at this point.

automata/fa/dfa.py
EduardoGoulart1: If the DFA is partial, transitions[state][letter] might not be defined. That's why I went with this one...

EduardoGoulart1:
> @EduardoGoulart1 @caleb531 in the latest push, I streamlined and update the test cases to incorporate partial DFAs, including tests that check for interoperability with partial and non-partial DFAs. This bought the line coverage up to 100% in the DFA file.
> 
> There are a couple of minor things left to do. The main thing is double checking over the test suite to ensure we are testing extensively enough for the partial behavior. The other thing is doing a pass over the DFA code to clean things up as much as possible. There were places where I removed a large amount of code, but I may have missed some things. Finally, we need to update the documentation.
> 
> Even though I'm not quite ready to call this done, I think this is in a mature enough state that we can start reviewing, as it should be mostly code complete at this point.

Does it make sense to start over. Now I am a bit lost in the middle of all the recent updates

eliotwrobson:
@EduardoGoulart1 I don't think it makes sense to make a new one, but you'll probably have to review the entire current set of changes instead of any one commit.

EduardoGoulart1:
Just had a quick look over it. I would say it's in a state to start reviewing it, so I will just drop my review...

eliotwrobson:
> But who is in charge of applying the changes?

Depending on the number of changes, I can probably apply them

tests/test_dfa.py
```diff
@@ -188,99 +269,42 @@ def test_read_input_step(self) -> None:
...
-            dfa <= other  # type: ignore
+            self.dfa <= other  # type: ignore
         with self.assertRaises(TypeError):
-            dfa > other  # type: ignore
+            self.dfa > other  # type: ignore
         with self.assertRaises(TypeError):
-            dfa >= other  # type: ignore
+            self.dfa >= other  # type: ignore
 
     def test_equivalence_not_equal(self) -> None:
```
EduardoGoulart1, line 291: This test should test the partial DFAs (applies to some other tests too). I was doing something like this:

```
def partial_combinations(*args):
    args = tuple(dfa.copy() for dfa in args)
    for partial_flags in product((False, True), repeat=len(args)):
        yield partial_flags, (
            to_partial(dfa, partial_flag)
            for partial_flag, dfa in zip(partial_flags, args)
        )
```

EduardoGoulart1: Then for each test I would just pass the DFAs to the `partial_combinations` function and apply the asserts normally

tests/test_dfa.py
```diff
@@ -838,72 +697,74 @@ def test_isfinite_finite(self) -> None:
...
+        self.assertNotEqual(dfa1, dfa2)
+
         # This DFA accepts all binary strings
         universal = DFA.universal_language(input_symbols)
         # This DFA represents the empty language
         empty = DFA.empty_language(input_symbols)
         # De Morgan's laws
-        self.assertEqual(~(contains_1111 | avoids_11), ~contains_1111 & ~avoids_11)
-        self.assertEqual(~(contains_1111 & avoids_11), ~contains_1111 | ~avoids_11)
+        self.assertEqual(~(dfa1 | dfa2), ~dfa1 & ~dfa2)
```
EduardoGoulart1, line 740: I think that here I prefer the original names... Makes it more understandable

EduardoGoulart1 (commented):
I am not so sure about the changes to the BFS method, otherwise minor changes

I think that the tests can be improved to try all combinations of the partial/complete DFAs with the code that I sent, but to be honest at this point I think that it's easier to just bring this PR to completion, so what we have is already good enough

automata/fa/dfa.py
```diff
@@ -702,26 +860,45 @@ def _find_state(
         return any(target_state_fn(state) for state in bfs_states)
 
     @staticmethod
-    def _cross_product(lhs: DFA, rhs: DFA) -> Tuple[DFAStateT, ExpandStateFn]:
-        """Builds the cross product between the two DFAs"""
+    def _cross_product(
```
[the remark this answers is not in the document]
eliotwrobson: I think the fix for this is actually pretty straightforward, just iterate through all input symbols in both inner dictionaries instead of the input symbol set each time in the inner loop.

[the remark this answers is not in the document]
EduardoGoulart1: This is already good enough, but would till be a problem if one of the DFAs is dense

automata/fa/nfa.py
```diff
@@ -201,6 +201,28 @@ def validate(self) -> None:
...
+        for input_symbol, next_states in chain.from_iterable(
+            self.transitions[state].items() for state in current_states
+        ):
+            # Ignore empty string and empty end states
+            if input_symbol and next_states:
+                input_symbol_set = res_dict.setdefault(input_symbol, set())
+                for end_state in next_states:
+                    input_symbol_set.update(lambda_closures[end_state])
+
+        for input_symbol, next_states in res_dict.items():
```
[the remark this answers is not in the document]
eliotwrobson: Would normally agree but the return type of this function is to match a specific signature somewhere else. Take a look at the from_nfa function.

automata/fa/dfa.py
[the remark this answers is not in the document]
EduardoGoulart1: Also minor comment: There are some TODOs introduced, I think they can be removed or we create an issue for them

automata/fa/dfa.py
```diff
@@ -1145,7 +1359,7 @@ def from_substring(
         # Computing failure function for partial matches as is done in the
         # Knuth-Morris-Pratt string algorithm so we can quickly compute the
         # next state from another state
-        kmp_table = [-1 for _ in substring]
+        kmp_table = array.array("i", (-1 for _ in substring))
```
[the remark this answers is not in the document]
eliotwrobson: Cheap way of saving memory since these are all integers.

EduardoGoulart1:
Since there are only minor comments from my side, and I don't want to block progress because of my limited time, I am good with whatever decisions you meet here so we can merge it (I can't approve my own PR)

tests/test_dfa.py
eliotwrobson: The original names are no longer descriptive. If you take a look at the parametrization of this function, it now tries all pairs of distinct DFAs from the parameters list instead of just two example DFAs.

This is actually the main test case checking for valid interop with complete and partial DFAs now.

automata/fa/dfa.py
```diff
@@ -218,10 +232,106 @@ def __iter__(self) -> Iterator[str]:
...
+        transitions: DFATransitionsT,
+        initial_state: DFAStateT,
+        final_states: AbstractSet[DFAStateT],
+        trap_state: DFAStateT,
+    ) -> Self:
+        added_trap_state = False
+        new_transitions: Dict[DFAStateT, Dict[str, DFAStateT]] = {
+            state: dict(state_path) for state, state_path in transitions.items()
+        }
+
```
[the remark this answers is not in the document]
eliotwrobson: I'll double check, but I think the old code always added a trap state. This function is for internal use only, and may get inputs that are already complete DFAs. Basically, the change in code is accompanying a change in the use case.

automata/fa/dfa.py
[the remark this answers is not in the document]
eliotwrobson: I think that efficiency issue is very specific to the intersection with dense DFAs. I'll leave a Todo that this can possibly be improved by writing a helper function to case on the booleans used as inputs, and we can implement this if anyone ever comes along with this issue.

As for the todos, they were mainly for myself as I didn't write tests for a couple of edge cases with partial DFAs. I'll remove those and add tests later today at some point.

automata/fa/dfa.py
```diff
@@ -1000,13 +1207,14 @@ def _populate_word_cache_up_to_len(self, k: int) -> None:
                     {
                         state: [
                             symbol + word
-                            for symbol in sorted_symbols
+                            for symbol in sorted_transition_symbols[state]
```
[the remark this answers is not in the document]
eliotwrobson: It could but it seems unnecessary, the only speed increase would come from removing dict lookups, and I'm not sure there's any real benefit to that.

automata/fa/dfa.py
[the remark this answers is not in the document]
caleb531: @eliotwrobson I'm fine with that, but can you please add a code comment explaining why `array.array` is used here?

tests/test_dfa.py
caleb531: @eliotwrobson I think the new names are fine, and agree with @eliotwrobson's reasoning for something more descriptive. Closing this thread.

automata/fa/dfa.py
caleb531: @EduardoGoulart1 Makes sense to be. Let's keep it the way it is.

automata/fa/nfa.py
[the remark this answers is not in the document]
caleb531: @EduardoGoulart1 Agreed with @eliotwrobson. `_iterate_through_symbol_path_pairs` needs to be a generator, meaning we need to `yield` the individual values instead of returning the entire iterable.

caleb531:
I've worked through a few of the comment threads, closing the ones that are taken care of, and commenting on a few others.

@EduardoGoulart1 @eliotwrobson Would it be possible for you guys to work through the outstanding comment threads? I would personally prefer to review this PR only once these comment threads have been resolved, and I don't have anything more to say on the ones that are still open.

eliotwrobson:
> Would it be possible for you guys to work through the outstanding comment threads? I would personally prefer to review this PR only once these comment threads have been resolved, and I don't have anything more to say on the ones that are still open.

@caleb531 sounds good, I'm resolving some now and the remaining threads are fairly short.

automata/fa/dfa.py
```diff
@@ -218,10 +232,106 @@ def __iter__(self) -> Iterator[str]:
...
+    # Supports partial (test)
+    def to_complete(self, trap_state: Optional[DFAStateT] = None) -> Self:
+        """
+        Turns a DFA (complete or not) into a complete DFA.
+        Might add trap state if the DFA does not have any
+        """
+        if not self.allow_partial:
+            return self.copy()
+
+        if trap_state is None:
```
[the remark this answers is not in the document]
eliotwrobson: Added check here with test in a later commit: https://github.com/caleb531/automata/pull/147/commits/85d8be7b4c8d584eb2f548062b3736959495381e

automata/fa/dfa.py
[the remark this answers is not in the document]
eliotwrobson: Use case changed slightly so I was able to switch back to the old code: https://github.com/caleb531/automata/pull/147/commits/b7fa44db7a80ab727d4324b389e3f8966e708c7e

automata/fa/dfa.py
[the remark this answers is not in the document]
eliotwrobson: Comment added here: https://github.com/caleb531/automata/pull/147/commits/b7fa44db7a80ab727d4324b389e3f8966e708c7e

automata/fa/dfa.py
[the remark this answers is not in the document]
eliotwrobson: Simplified logic and added comment per discussion: https://github.com/caleb531/automata/pull/147/commits/85d8be7b4c8d584eb2f548062b3736959495381e

automata/fa/fa.py
```diff
@@ -67,6 +67,7 @@ def iter_transitions(self) -> Generator[Tuple[FAStateT, FAStateT, str], None, No
         of the form (from_state, to_state, symbol)
         """
 
+    # Supports partial DFA
```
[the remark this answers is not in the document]
eliotwrobson: Removed here https://github.com/caleb531/automata/pull/147/commits/2a72416dc1a16d120c9be0c4e718ca843865003e

tests/test_dfa.py
eliotwrobson: This is subsumed by the `test_set_laws` function, where the equality function is used on a mix of partial and complete DFAs.

automata/fa/dfa.py
eliotwrobson: Sorry, I was a bit unclear in my original comment. It makes sense why this line was changed, but we can call `keys()` instead of `items()` since the `tgt_state` variable isn't being read. I'll update the code with this change.

eliotwrobson:
@caleb531 all items are resolved and removed all TODOs. The only thing left is docs changes, but I'll leave that until after your review. Thanks!

automata/fa/dfa.py
```diff
@@ -39,9 +41,11 @@
 DFAPathT = Mapping[DFASymbolT, DFAStateT]
 DFATransitionsT = Mapping[DFAStateT, DFAPathT]
 
-ExpandStateFn = Callable[[DFAStateT], Iterator[Tuple[DFASymbolT, DFAStateT]]]
+ExpandStateReturnType = Iterator[Tuple[DFASymbolT, DFAStateT]]
+ExpandStateFn = Callable[[DFAStateT], ExpandStateReturnType]
 IsFinalStateFn = Callable[[DFAStateT], bool]
 TargetStateFn = Callable[[DFAStateT], bool]
+BooleanLiteral = Literal[True, False]
```
caleb531, line 48: @eliotwrobson @EduardoGoulart1 I believe `BooleanLiteral` is unnecessary. You can simply use `bool` as the type hint (e.g. `lhs_relevant: bool` instead of `lhs_relevant: BooleanLiteral`).

automata/fa/dfa.py
```diff
@@ -438,17 +539,21 @@ def _minify(
...
             name: {
                 letter: back_map[transitions[next(iter(eq))][letter]]
-                for letter in input_symbols
+                for letter in transitions[next(iter(eq))].keys()
             }
             for name, eq in eq_class_name_pairs
         }
 
+        allow_partial = any(
+            len(lookup) != len(input_symbols) for lookup in transitions.values()
```
caleb531, line 548: @EduardoGoulart1 @eliotwrobson I think a DFA can also be considered partial if the top-level transitions dict doesn't include all of the states as its keys (i.e. any states not listed are trap states). For example:

```python
# A DFA whose language is just {"a"}
dfa = DFA(
    states={"q0", "q1"},
    input_symbols={"a", "b"},
    transitions={
        # This case is already covered in the above check, since the transition to "b" is left out
        "q0": {"a": "q1"}
        # Definition for "q1" is omitted because it has no transitions anyway;
        # this case is not considered in the above check to determine a DFA"s partiality
    },
    initial_state="q0",
    final_states={"q1"},
    allow_partial=True,
)
```

Please correct me if I misunderstand the code here, but it just caught my eye.

automata/fa/dfa.py
```diff
@@ -702,25 +830,49 @@ def _find_state(
         return any(target_state_fn(state) for state in bfs_states)
 
     @staticmethod
-    def _cross_product(lhs: DFA, rhs: DFA) -> Tuple[DFAStateT, ExpandStateFn]:
-        """Builds the cross product between the two DFAs"""
+    def _cross_product(
+        lhs: DFA, rhs: DFA, lhs_relevant: BooleanLiteral, rhs_relevant: BooleanLiteral
```
caleb531, line 834: @EduardoGoulart1 @eliotwrobson For clarity, I request that the two new parameters at the end be keyword-only. And, as I mentioned in my above comment, you should be able to just use `bool` for the type hints. That is:

```python
def _cross_product(lhs: DFA, rhs: DFA, *, lhs_relevant: bool, rhs_relevant: bool) -> Tuple[DFAStateT, ExpandStateFn]:
    # ...
```

Even though the method is internal, this will make it much clearer later in the code where you are currently passing `True, True` to this method.

automata/fa/dfa.py
```diff
@@ -222,6 +226,102 @@ def __len__(self) -> int:
         """Returns the cardinality of the language represented by the DFA."""
         return self.cardinality()
 
+    def to_partial(self) -> Self:
```
caleb531, line 229: @EduardoGoulart1 @eliotwrobson This new method needs to be added to the API Documentation (in `docs/fa/class-dfa.md`).

caleb531 (changes requested):
@EduardoGoulart1 @eliotwrobson Left a few requested changes. Nothing major (at least from my perspective 😅).

automata/fa/dfa.py
eliotwrobson: @caleb531 agreed, done here! https://github.com/caleb531/automata/pull/147/commits/070ff3206873feee4728c796792b87ff9d88914f

automata/fa/dfa.py
eliotwrobson: @caleb531 this functionality was removed here, take a look at the commit message: https://github.com/EduardoGoulart1/automata/commit/43f377c4b7aa6381c286679a0379776857e2459c

I don't think it's worth adding support for this, since it would lead to some incompatibilities with existing code, and we'd have to go back and test some extra things. This only leads to minor memory savings in very limited circumstances anyway.

caleb531: @eliotwrobson Okay, fair enough. The current API documentation doesn't say anything against your recommended behavior anyway

> `allow_partial`: by default, each DFA state must have a transition to
every input symbol; if `allow_partial` is `True`, you can disable this
characteristic (such that any DFA state can have fewer transitions than input
symbols)

automata/fa/dfa.py
eliotwrobson: @caleb531 updated here!

caleb531: @eliotwrobson Yep, just reviewed it! Looks great to me!

caleb531 (approved):

eliotwrobson (approved):
@caleb531 Resolved all threads from your comments! If those were all the items you had questions about, feel free to merge when ready!

caleb531:
@EduardoGoulart1 @eliotwrobson This looks good to me! Will merge now.

PR #149, opened 2023-05-13
Transition table example

eliotwrobson:
Add an example of creating a transition table, adapted from visual automata. This is part of the deprecation of visual automata in #124, as it was decided that we did not want to add pandas as a dependency for this library. Thus, the best course of action was to add snippets of code from that library that would have introduced dependencies as examples.

Should wait for #129 before merging, as this code uses a method introduced there.

@lewiuberg please let me know what you think if you get the chance.

### DIFF
--- a/docs/fa/examples.md
+++ b/docs/fa/examples.md
@@ -143,6 +143,10 @@ def make_table(target_fa) -> pd.DataFrame:
         if to_state in final_states:
             to_state_str = "*" + to_state_str
 
+        # Prepare nice symbol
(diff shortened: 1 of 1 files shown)


caleb531:
@eliotwrobson This looks great! But I'm gonna switch this PR to Draft mode since I don't want to accidentally merge it before #129. 😅

eliotwrobson:
Now that #129 has been merged, the example code added here should work! Going to go ahead and merge.

PR #152, opened 2023-06-17
Add word cache reset function

### LINKED ISSUE #148: v8 Caching Behavior
Followup discussion from this thread: https://github.com/caleb531/automata/pull/147#discussion_r1190404517

@EduardoGoulart1 What are your thoughts on having an enum with the cache levels given in this discussion?

Re: the paper you linked, that's the algorithm I was referring to. That algorithm has complexity depending on the number of transitions, so for a partial DFA over a large alphabet, that algorithm should be substantially faster.

eliotwrobson:
See title, adds a word cache reset function and precomputes the digraph used by the DFA. Resolves #148 

In looking at this, I thought about using a cached property for some of the methods that return fixed information about the given DFAs. However, switching these to cached properties (with the decorator) requires using a slightly different API. @caleb531 what are your thoughts about changing this? Is there a way to provide this caching without changing the API? It would be great if there were a way to do this cleanly.

### DIFF
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -63,6 +63,9 @@ class Automaton(metaclass=abc.ABCMeta):
         #   because __setattr__ is disabled due to immutability
         self.__init__(**d)  # type: ignore
 
+    def __contains__(self, item: str) -> bool:
+        return self.accepts_input(item)
+
     @abc.abstractmethod
     def read_input_stepwise(self, input_str: str) -> Generator[Any, None, None]:
         """Return a generator that yields each step while reading input."""
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -63,9 +63,6 @@ class Automaton(metaclass=abc.ABCMeta):
         #   because __setattr__ is disabled due to immutability
         self.__init__(**d)  # type: ignore
 
-    def __contains__(self, item: str) -> bool:
-        return self.accepts_input(item)
-
     @abc.abstractmethod
     def read_input_stepwise(self, input_str: str) -> Generator[Any, None, None]:
         """Return a generator that yields each step while reading input."""
--- a/automata/base/automaton.py
+++ b/automata/base/automaton.py
@@ -151,6 +151,10 @@ class Automaton(metaclass=abc.ABCMeta):
         )
         return f"{self.__class__.__qualname__}({values})"
 
-    def __contains__(self, input_str: str) -> bool:
+    def __contains__(self, item: Any) -> bool:
         """Returns whether the word is accepted by the automaton."""
-        return self.accepts_input(input_str)
+
+        if not isinstance(item, str):
+            return False
+
+        return self.accepts_input(item)
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -84,9 +84,9 @@ class DFA(fa.FA):
             _digraph=self._get_digraph(transitions),
         )
 
-        self.reset_word_cache()
+        self.reset_cache()
 
-    def reset_word_cache(self) -> None:
+    def reset_cache(self) -> None:
         """Resets the word and count caches. Can be called if too much memory is being used."""
         object.__setattr__(self, "_word_cache", [])
         object.__setattr__(self, "_count_cache", [])
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -62,6 +62,7 @@ class DFA(fa.FA):
     allow_partial: bool
     _word_cache: List[DefaultDict[DFAStateT, List[str]]]
     _count_cache: List[DefaultDict[DFAStateT, int]]
+    _digraph: nx.DiGraph
 
     def __init__(
         self,
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -56,6 +56,7 @@ class DFA(fa.FA):
         "allow_partial",
         "_count_cache",
         "_word_cache",
+        "_digraph",
     )
 
     allow_partial: bool
@@ -80,9 +81,21 @@ class DFA(fa.FA):
             initial_state=initial_state,
             final_states=final_states,
             allow_partial=allow_partial,
+            _digraph=self._get_digraph(transitions),
         )
         object.__setattr__(self, "_word_cache", [])
         object.__setattr__(self, "_count_cache", [])
+
+    @staticmethod
+    def _get_digraph(transitions: DFATransitionsT) -> nx.DiGraph:
+        """Return a digraph corresponding to this DFA with transition symbols ignored"""
+        return nx.DiGraph(
+            [
+                (start_state, end_state)
+                for start_state, transition in transitions.items()
+                for end_state in transition.values()
+            ]
+        )
 
     def __eq__(self, other: Any) -> bool:
         """
@@ -312,16 +325,6 @@ class DFA(fa.FA):
 
         if not ignore_rejection:
             self._check_for_input_rejection(current_state)
-
-    def _get_digraph(self) -> nx.DiGraph:
-        """Return a digraph corresponding to this DFA with transition symbols ignored"""
-        return nx.DiGraph(
-            [
-                (start_state, end_state)
-                for start_state, transition in self.transitions.items()
-                for end_state in transition.values()
-            ]
-        )
 
     def minify(self, retain_names: bool = False) -> Self:
         """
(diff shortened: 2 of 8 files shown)


docs/fa/class-dfa.md
```diff
@@ -309,6 +309,14 @@ for word in dfa:
     print(word)
 ```
 
+## DFA.reset_cache(self)
+
+Resets the word and count caches. Can be called if too much memory is being used.
+
+```python
+dfa.reset_word_cache()
```
caleb531, line 317: @eliotwrobson Just a note: this name does not match the actual method name in the `DFA` class.

tests/test_dfa.py
```diff
@@ -2504,3 +2504,20 @@ def test_empty_language(self) -> None:
 
         dfa = DFA.empty_language({"0", "1", "a", "b"})
         self.assertTrue(dfa.isempty())
+
+    def test_reset_word_cache(self) -> None:
+        max_len = 4
+        dfa = DFA.of_length(set("01"), min_length=0, max_length=max_len)
```
caleb531, line 2510: @eliotwrobson This is totally a trivial nitpick, but you can use a set literal here instead of the `set()` function (i.e. `{"0", "1"}`). This is how it's written for the other `DFA.of_length` calls in the codebase.

automata/fa/dfa.py
```diff
@@ -80,10 +82,30 @@ def __init__(
             initial_state=initial_state,
             final_states=final_states,
             allow_partial=allow_partial,
+            _digraph=self._get_digraph(transitions),
         )
+
+        self.reset_cache()
+
+    def reset_cache(self) -> None:
```
caleb531, line 90: @eliotwrobson I would kindly ask that the method be called `clear_cache()` rather than `reset_cache()`. I think the former has better semantics and is the more common verb to use in a caching context.

This requested change would include updating the verbiage in the docstrings, Markdown docs, and test function names for consistency, as well.

caleb531 (changes requested):
@eliotwrobson Overall looks good, but left you a few comments with some requested changes (mostly the name of the method itself).

Regarding preserving the API, I believe the [@functools.lru_cache()](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator is the stdlib way to cache the output of methods. The only thing to note is that the method returns the cached output based on the arguments you pass to the function at call time. So `foo(first='Eliot')` and `foo(last='Robson')` will each be cache misses on the first call. But I think this is the behavior we want in most cases, anyway.

eliotwrobson:
@caleb531 sounds good! As you pointed out, I made a mistake in the docs with the name in the first place 😅 so changing it is no problem.

Regarding caching, using lru cache here is actually a mistake: https://youtu.be/sVjtp6tGo0g

The use here would be for caching something like `isempty`, where there's no sense in recomputing the result each time for an immutable DFA. The canonical way of doing this as of 3.8 is with this decorator: https://docs.python.org/3/library/functools.html#functools.cached_property

However, this changes the method into a property (so it is accessed by the client code differently). This might be more natural since users may expect cached properties to behave this way, but I also don't know if changing the API this way is the right move.

caleb531:
@eliotwrobson Ah yes, I forgot that `lru_cache` is problematic for methods (ironically, this is something I myself [have mentioned in a previous issue](https://github.com/caleb531/automata/pull/80#issuecomment-1289882298), yet have apparently forgotten 😅). However, it seems that there are at least two workarounds:

1. Implement the [weakref-based decorator for caching methods](https://stackoverflow.com/a/33672499/560642) (if we go this approach, I would call this decorator `cached_method` instead of `memoized_method` to achieve parity with the below package, plus I just like the name better)
2. Require the [`cached_method` package](https://pypi.org/project/cached_method/) as an additional dependency

The package doesn't seem like much overhead, but given how small this decorator is, we could just as easily include the implementation directly in the code. I could go either way.

So this whole post of mine obviously implies that I'd still prefer members like `isempty` to be methods, not properties. For me, it isn't just about preserving API backwards-compatibility, but about the semantics of methods vs. properties and the flexibility that methods provide.

Methods can have additional arguments in a way that's effectively future-proof with keyword arguments; properties can't no matter what you do. And as for the semantic argument: I tend to use properties to represent nouns/things, but when I have some check or operation to perform, I use a method.

I recognize that `Automaton.input_parameters` is the only exception to my rule of thumb because it represents a thing (or _things_), not an operation. Plus, the code is pretty simple and hardly expensive.

Okay, thank you for coming to my TED talk. Please let me know your thoughts!

eliotwrobson:
I tend to agree, having these be methods allows more flexibility and conveys these are something to be computed. In the interest of writing less code, I'm inclined to go with the adding the optional dependency. So I will do that along with the renames described.

automata/fa/dfa.py
eliotwrobson: Done here: https://github.com/caleb531/automata/pull/152/commits/61e16a660cdd8bb455adf081b0bc4e9d44fb03ac

docs/fa/class-dfa.md
eliotwrobson: Fixed here: https://github.com/caleb531/automata/pull/152/commits/61e16a660cdd8bb455adf081b0bc4e9d44fb03ac

tests/test_dfa.py
eliotwrobson: Done here https://github.com/caleb531/automata/pull/152/commits/8ba372fcfdbe58e87fd5128113f13e5ca889903f

automata/fa/dfa.py
```diff
@@ -56,6 +57,9 @@ class DFA(fa.FA):
         "allow_partial",
         "_count_cache",
         "_word_cache",
+        # These two entries are to allow for caching methods
+        "__dict__",
```
caleb531, line 61: @eliotwrobson I always thought `__dict__` and `__slots__` were not something you use together. Do we really need to add it here?

automata/base/automaton.py
```diff
@@ -63,9 +63,6 @@ def __setstate__(self, d: Dict[str, Any]) -> None:
         #   because __setattr__ is disabled due to immutability
         self.__init__(**d)  # type: ignore
 
-    def __contains__(self, item: str) -> bool:
```
caleb531: @eliotwrobson I was gonna say—there's already an `Automaton.__contains__` method defined around like 154.

eliotwrobson: Yes, was working a little bit too fast on these changes 😅

automata/fa/dfa.py
eliotwrobson: We do need to add `__dict__` here (removing it causes the caching to fail). I think we still get speed benefits from slots, however.

caleb531:
@eliotwrobson Just want to say please feel free to continue to make changes, and just convert this PR out of Draft mode when it's ready for my re-review.

automata/fa/dfa.py
caleb531: @eliotwrobson Sweet! Works for me.

automata/base/automaton.py
```diff
@@ -151,6 +151,10 @@ def __repr__(self) -> str:
         )
         return f"{self.__class__.__qualname__}({values})"
 
-    def __contains__(self, input_str: str) -> bool:
+    def __contains__(self, item: Any) -> bool:
```
eliotwrobson, line 154: Changed this because mypy was complaining. Like some of the other dunder methods, I think this has to accept anything as a possible argument.

caleb531: @eliotwrobson That's fine! But if we're gonna do that, I wonder if we should have some sort of type check here to ensure that the item is an iterable or something. Or maybe it's not even worth the trouble.

eliotwrobson:
@caleb531 looking over the changes I think this is ready to go! Flipping to open.

automata/base/automaton.py
eliotwrobson: Oh in the changes I made to this function, I added an `isinstance` check that returns false if anything other than a string is passed in.

caleb531: @eliotwrobson Oh perfect, I see that now. Excellent, thank you!

caleb531 (approved):
@eliotwrobson LGTM! 👍

caleb531:
Actually, one quick afterthought:

How will all this additional caching behave if a user enables mutable automata? I guess they can just clear the caches manually as they make changes, correct?

eliotwrobson:
> How will all this additional caching behave if a user enables mutable automata? I guess they can just clear the caches manually as they make changes, correct?

@caleb531 yes, they can manually clear cached methods by emptying the `__dict__`. I don't think people will run into issues (as the API still assumes immutability even though the data structures themselves don't prevent this behavior).

PR #155, opened 2023-06-20
Add table property to fa

lewiuberg:
Add FA transition table as a property.

### DIFF
--- a/automata/fa/fa.py
+++ b/automata/fa/fa.py
@@ -8,9 +8,20 @@ import pathlib
 import random
 import uuid
 from collections import defaultdict
-from typing import Dict, Generator, List, Literal, Optional, Set, Tuple, Union
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    List,
+    Literal,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 
 from automata.base.automaton import Automaton, AutomatonStateT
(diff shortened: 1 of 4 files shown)


eliotwrobson:
@lewiuberg This isn't quite what I had in mind in the discussion on #124 😅 Basically, we didn't want to add the table and input check functions because of styling issues and having to add Pandas as a dependency. So instead, we decided it would be best to include the code doing this as part of the documentation (so people can copy it) and not directly as code in the library. I was able to convert the table functionality in #149, and you mentioned that you found `input_check` useful, so the idea was to add `input_check` to the documentation as well.

Either way, thank you for the work you put into this PR! But will close as this is essentially the same as what was added in #149.

lewiuberg:
Ah! I see! Then I won't have much to contribute on this. Adding pandas as a dependency is one of the libraries I worry least about 😊

lewiuberg:
I hope I didn't come of the wrong way here. While I'm a bit more "liberal" when it comes to dependencies I can see it from your perspective too. The make_table function you have included is a good option. Would you be open to me making a draft version of docs in mkdocks for this project? Something like one I am working on now: https://uberg.me/setup/

eliotwrobson:
@lewiuberg no worries! And absolutely, I think having more accessible documentation would be excellent. Maybe just touch base with @caleb531 before starting.

PR #163, opened 2023-07-19
Add JOSS paper

eliotwrobson:
See title @caleb531


.github/workflows/draft-pdf.yml
```diff
@@ -0,0 +1,23 @@
+on: [push]
```
caleb531, line 1: @eliotwrobson So we don't need to be uploading the paper on every single commit, do you think it's reasonable to limit this action to run only when files in the `joss` directory change? [This StackOverflow post](https://stackoverflow.com/questions/63822219/how-to-run-github-actions-workflow-only-if-the-pushed-files-are-in-a-specific-fo) indicates that GitHub Actions supports this out of the box with [path filters](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore).

I believe in our case, the syntax would be:

```yml
on:
  push:
    paths: ["joss/**"]
```

eliotwrobson: Seems reasonable! will add.

joss/paper.md
```diff
@@ -0,0 +1,151 @@
...
+included functionality for parsing regular expressions and manipulating finite-state
+machines enables fast and accessible exploration of these structures by researchers.
+On the educational side, the package includes visualization logic that allows students to
+interact with these structures in an exploratory manner, and has already seen usage in
+undergraduate courses. `automata` has already been cited in publications [@Erickson23], with more
+to come as the package becomes more popular.
+
+`automata` has seen a large number of contributions by external contributors and wide adoption,
+demonstrating the demand for a high-quality Python package providing these features. The code is
+well maintained, including a comprehensive test suite and type annotations, meaning new features
```
caleb531, line 84: @eliotwrobson If I'm not mistaken, I think "well maintained" needs to be hyphenated (i.e. "well-maintained").

(the academic context of this paper is my only excuse to nitpick the grammar 😅)

.github/workflows/draft-pdf.yml
eliotwrobson: Done here: https://github.com/caleb531/automata/pull/163/commits/ef46c654e4d32e72519fe31bbf85a46121a0f737

joss/paper.md
eliotwrobson: Good catch! Done here: https://github.com/caleb531/automata/pull/163/commits/ef46c654e4d32e72519fe31bbf85a46121a0f737

joss/paper.md
```diff
@@ -0,0 +1,151 @@
...
+[dengl11](https://github.com/dengl11),
+[Tagl](https://github.com/Tagl),
+[lewiuberg](https://github.com/lewiuberg),
+[CamiloMartinezM](https://github.com/CamiloMartinezM),
+[abhinavsinha‑adrino](https://github.com/abhinavsinha-adrino),
+[EduardoGoulart1](https://github.com/EduardoGoulart1), and
+[khoda81](https://github.com/khoda81)
+for their invaluable code contributions to this project.
+
+# References
```
caleb531, line 151: @eliotwrobson Is it fine to leave the _References_ section blank?

eliotwrobson: This is automatically populated during the build.

.github/workflows/draft-pdf.yml
caleb531: @eliotwrobson Good call for adding the `pull_request` section too. Closing this thread.

caleb531:
@eliotwrobson Thanks for working through my requested changes and questions! I've reviewed the entire paper, and content-wise, it looks great! There is one other issue I identified, though (not with the paper itself):

The checks on this PR indicate that the build workflow is failing (the one I just added today to build the wheel on every commit). The reason appears to be because the build process thinks `joss` is a top-level Python package 🤷‍♂️.

```
error: Multiple top-level packages discovered in a flat-layout: ['joss', 'automata'].
```

Logs: https://github.com/caleb531/automata/actions/runs/5594307830/jobs/10229019972?pr=163#step:7:13

I'm at a loss for why `joss` is considered a Python package (and just to clarify, not "package" like PyPI package, but rather "package" as in [a container of Python modules](https://docs.python.org/3/tutorial/modules.html#packages)).

Maybe the code from [this SO post](https://stackoverflow.com/questions/72294299/multiple-top-level-packages-discovered-in-a-flat-layout) would do the trick?

```toml
[tool.setuptools]
py-modules = ['automata']
```

I know the SO post suggests leaving the array empty, but I would prefer explicit over implicit if my tweaked version still works.

eliotwrobson:
> I'm at a loss for why `joss` is considered a Python package

@caleb531 I think that this has to do with the fact that Python pulls its module structure from the directory structure. So even with a directory without any Python files, it's treating it like a module (for whatever reason).

Either way, I updated the `pyproject.toml` file as you suggested and the build passed! https://github.com/caleb531/automata/pull/163/commits/86b5c5bf3e91505b8ea687fb5ff3f6ff443785c2

If everything looks good now, feel free to approve + merge + complete the paper submission form.

caleb531:
@eliotwrobson Unfortunately, we may need to update the build job to be a little smarter. If you build the wheel locally:

```sh
python -m build --sdist --wheel --outdir dist/ .
```

Look inside the `dist/` directory that gets created, and open the `.tar.gz` file (which represents the source distribution). Most, if not all, of the `automata` modules are missing.

I'm sorry to say, but I'd prefer not to merge this PR until we get this resolved so that `develop` remains functional. I'm looking into alternative configuration now to see if there's a better solution than `my-modules`—I'll keep you posted.

caleb531:
@eliotwrobson Fixed it! Ended up being super simple. See my newly-pushed commit.

Basically, needed to be:

```
# Only one module in this repo
[tool.setuptools.packages.find]
exclude = ["joss"]
```

caleb531 (approved):

caleb531:
@eliotwrobson I've approved/merged the PR now! Please let me know what are the next steps to formally publish to JOSS.

eliotwrobson:
@caleb531 awesome! The last step is just filling out the form here: https://joss.theoj.org/papers/new

The form is very short. You can double-check if you would like, but our paper meets all of the requirements.

caleb531:
@eliotwrobson Gotcha. This is what I'm planning to submit:

### Title
automata: A Python package for simulating and manipulating automata

## Software's Git repository URL

https://github.com/caleb531/automata

### Name of git branch containing the paper, not the path

develop

### Software version

v8.0.0

### Type of submission

New Submission

### Main subject of the paper

_I'd prefer if you write this, if you can, please_

### Message to editors

_Guessing we don't need this?_

eliotwrobson:
@caleb531 all looks good! For main subject, I think this is just a keyword, so "automata" is fine. And the message to the editors is just supposed to be a short description of the software in a couple of sentences. Pulled partially from the README.md, you can just put
```
Automata is a Python 3 library that implements the structures and algorithms for finite automata, pushdown automata, and Turing machines. The algorithms have been optimized for good performance with large workloads. No portion of the software has been directly submitted for publication previously, but it has been cited in previous works for use as educational software.
```

caleb531:
@eliotwrobson Gotcha! Although after some further thought, I am actually comfortable with you submitting the paper on behalf of the project. I think you understand the process better, and would be better suited for any future resubmissions, etc.

Only if that's fine with you, though. But I personally think you would be better suited all-around.

eliotwrobson:
@caleb531 sure! I'll go ahead and submit.

PR #164, opened 2023-07-19
Add internal minify call to to_partial function

eliotwrobson:
Adds an optional (defaulted to on) minify step to the `to_partial` function. Required some small changes to the `_minify` logic so it wouldn't complain about missing end states in the transition dictionary passed in.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -224,7 +224,7 @@ class DFA(fa.FA):
         """Returns the cardinality of the language represented by the DFA."""
         return self.cardinality()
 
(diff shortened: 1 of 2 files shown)


caleb531 (approved):
This all looks good to me! 👍

eliotwrobson:
@caleb531 great! I let this sit just because the semantics were a tiny bit wonky (we might be overly aggressive with the minification), but I think overall it's good to continuously minify to prevent DFA blowup from a large number of operations. In the worst case, we can just change the semantics to be more coherent later (I expected some rough edges with the greater partial support anyway).

PR #167, opened 2023-08-15
Aho-Corasick

### LINKED ISSUE #162: Aho-Corasick Algorithm
See title, implement the Aho-Corasick algorithm for construction of a DFA which recognizes all strings which contain a substring from a given dictionary. This is an extension of the existing `from_substring` method that uses the KMP construction algorithm (and can probably replace that method if implemented, as the existing method is just a special case of the new one).

References:
- https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm
- https://www.toptal.com/algorithms/aho-corasick-algorithm
- https://cp-algorithms.com/string/aho_corasick.html#construction-of-an-automaton

cc @Tagl as you originally implemented the KMP algorithm. I'm not sure if there's a simple way to extend the existing implementation to the more general AC algorithm.

Tagl:
Closes #162 
@caleb531 @eliotwrobson 
Haven't added tests for full coverage but I've done a fair bit of manual testing and it is producing correct results.
The DFA produced isn't necessarily minimal (consider {"ab", "abba"}) but it's certainly faster than taking multiple unions.
I've been trying to cut it down to a single BFS and then remove the object oriented nodes but there's always something that just isn't clicking for me, so I'm keeping it like this for now, maybe you will spot it.
I have to put this on hold on my end for a few days.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1857,7 +1857,7 @@ class DFA(fa.FA):
         If must_be_suffix is set to True, then the each substring must be a suffix instead.
         """
         class OutNode:
-            def __init__(self, keyword: str, next_node: OutNode) -> None:
+            def __init__(self, keyword: str, next_node: Optional[OutNode]) -> None:
                 self.keyword: str = keyword
                 self.successor: Optional[OutNode] = next_node
 
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1751,7 +1751,7 @@ class DFA(fa.FA):
     def from_suffixes(
         cls: Type[Self],
         input_symbols: AbstractSet[str],
-        suffixes: str,
+        suffixes: AbstractSet[str],
         *,
         contains: bool = True,
     ) -> Self:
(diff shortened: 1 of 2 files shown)


automata/fa/dfa.py
```diff
@@ -1347,6 +1364,106 @@ def from_substring(
...
+        contains: bool = True,
+        must_be_suffix: bool = False,
+    ) -> Self:
+        """
+        Directly computes a DFA recognizing strings containing at least one of the given substring.
+        The implementation is based on the Aho-Corasick string-searching algorithm.
+        If contains is set to False then the complement is constructed instead.
+        If must_be_suffix is set to True, then the each substring must be a suffix instead.
+        """
+        class OutNode:
```
eliotwrobson, line 1382: Will take a closer look later, but I think some of these node classes can be simplified by using the strings themselves as state names? I believe this is constructing a trie just like in `from_finite_language`, maybe look there for ideas (assuming I'm correctly understanding how you want to simplify this implementation).

Tagl: Using the strings as state names is neither time nor memory efficient though. I would rather use integers myself. The keyword string is honestly useless in our case, it is a relic of the C++ implementation so it can be removed. At that point the OutNode is just a singly linked list which can surely be represented differently.

eliotwrobson: You raise some good points. I just got back from traveling so I'll try to take a closer look, but in general I agree that I think there's some way the complexity of this implementation can be reduced somewhat.

Tagl: We can put it in like this and then try to reduce complexity later. It's still rather useful.
We can also add another utility function which makes use of it, constructing a DFA forbidding or requiring consecutive repetitions of identical characters.

Tagl: I think I found a bug, will try to make a small test case that produces it.

eliotwrobson: Sounds good and no rush! Right now we're focusing on the docs changes in anticipation of the PyOpenSci submission. As that's the priority, we probably won't merge this until we get the ball rolling on that.

Tagl: There was no bug, I simply made a human error.
Implementation is fine and I have field tested it rather thoroughly.
Performance wise it has been alright so far, haven't benchmarked anything but has worked for my use case of up to 100000 strings of length in [0, 32] and a few other tests of various sizes.

eliotwrobson: Ahh that makes sense. As long as the efficiency is reasonable, I'm ok with merging as-is and we can speed things up later if people run into issues. The main thing we need to add to this PR is a docstring, but otherwise the code is relatively clean and the test cases look good @Tagl

Tagl: What is missing from the docstring?

eliotwrobson: If you look at the main branch, all the public functions have numpy style docstrings that describe the arguments and return value. The new documentation site is generated using these: https://github.com/caleb531/automata/blob/main/automata/fa/dfa.py#L714

Tagl: Sure, the algorithm does not need changes, but I won't be free to work on this until Friday at the earliest

eliotwrobson (approved):
Just rebased with a new docstring, everything looks good! Will merge soon.

@Tagl I wound up getting rid of the from_suffixes function, just because the API for this class is getting a little bloated, and that function was only a passthrough for the new functionality here.

PR #169, opened 2023-09-03
Address paper review

eliotwrobson:
See title, dummy PR so I can review changes and squash + merge to reduce the number of commits to develop.


joss/paper.bib
```diff
@@ -17,7 +17,7 @@ @Misc{brics
   author =  {Anders M\o{}ller},
   title =   {dk.brics.automaton -- Finite-State Automata
              and Regular Expressions for {Java}},
-  note =    {\texttt{http://www.brics.dk/automaton/}},
+  url =    {http://www.brics.dk/automaton/},
```
caleb531, line 20: @eliotwrobson Shouldn't this be an HTTPS URL?

eliotwrobson: I'm not quite sure. This URL is taken from the citation given by the website of the other library, but that could contain a mistake.

caleb531: @eliotwrobson Well, my thought is that since this isn't a URL to a schema or XML namespace (where a plain HTTP URL is historically required; [see this SO post](https://stackoverflow.com/questions/30707609/xml-namespace-uri-with-https)), I would think it would be fine to use HTTPS. Especially when I visit the URL, I am taken to a real webpage with content.

eliotwrobson: That makes sense to me, I can make the change when I'm back at my desk.

PR #170, opened 2023-09-08
Add pairwise helper function

eliotwrobson:
See title. Adds pairwise helper function and simplifies some pairwise iterations (avoids making unnecessary copies of lists). Mostly a style thing.


caleb531 (approved):
Makes sense to me! Approved.

Issue #159, opened 2023-06-27
Cython for optimizing certain computations


eliotwrobson:
Creating this issue because I've been looking at Cython a bit, and wanted to record my thoughts on how it might enhance this project. As a starting point, vanilla Python code can be compiled with Cython and should run faster than just using Python, and this optimization can be further enhanced without having to dive too deep into non-Pythonic constructs. These optimizations should be especially useful for work-intensive loops (such as the `refine` method on the `PartitionRefinement` data structure). So adding Cython should substantially benefit this library without having to rewrite large amounts of code. Based on some previous profiling done in #95, there are a few intensive computations we could target and extract into a Cython file, leaving everything else as Python.

The main obstacle is that Cython needs to be complied, and this requires adding some configuration to the build step (and may also require reconfiguring the way tests get run). I found an example [here](https://levelup.gitconnected.com/how-to-deploy-a-cython-package-to-pypi-8217a6581f09) and [here](https://github.com/FedericoStra/cython-package-example). We should be able to keep everything really simple to start (we only really need a single file that will contain helper functions for expensive computations used by the DFA class).

I think this is probably too much for the v8 release (as that's just waiting on finishing up testing and implementation of one more feature), but it would be cool to target for v9. Aside from setting up the project properly, I don't think there should be too much work involved, as this would just be converting code that we have already with some optimizations.

Any thoughts? @caleb531 (cc @Tagl, as you did the original profiling)

Tagl:
I've never used Cython myself so I'm unaware how much of a difference it makes.
I fully support optimizing however we can and from profiling I could see the refine method being the main bottleneck in my use case.

eliotwrobson:
Got a very minimal version of this working with light optimizations (based on length of translated code). Haven't benchmarked but seems to be faster: https://github.com/eliotwrobson/automata/tree/cython

eliotwrobson:
@caleb531 

> From a maintenance perspective on our end, I personally don't have a problem with a build step. I would only hope that the building of the project would happen automatically when users install the library via pip (seems like your `setup.py` would take care of this).

[...]

PR #177, opened 2023-10-13
Added `show_diagram` method for Push Down Automata

Vipul-Cariappa:
Implemented the `show_diagram` method for Push Down Automata.
I have just adapted the `FA.show_diagram` (Finite Automata) code for `PDA.show_diagram`.

I have not yet implemented `show_diagram` with `input_str`. I will require some suggestions here. Should I also show the operations performed on the stack, or only the state transitions are sufficient?

I have not tested the code, except for `DPDA` and `NPDA` in the docs. I have attached the images below.

DPDA:
image

NPDA:
image

Reminder for myself and the maintainer: I should also update the docs to reflect the new feature, before merging this PR.

Hope this is helpful. Please let me know if any changes are required.

---

Edited:

I have implemented the construction of the diagram with an `input_str`.

DPDA from docs with `input_str = "aabb"`:
image

NPDA from docs with `input_str = "abba"`:
image

I have updated the docs. I should now write down some test cases.

### DIFF
--- a/automata/base/exceptions.py
+++ b/automata/base/exceptions.py
@@ -87,3 +87,9 @@ class InfiniteLanguageException(AutomatonException):
     """The operation cannot be performed because the language is infinite"""
 
     pass
+
+
+class DiagramException(Exception):
+    """The diagram cannot be produced"""
+
+    pass
--- a/automata/base/exceptions.py
+++ b/automata/base/exceptions.py
@@ -89,7 +89,7 @@ class InfiniteLanguageException(AutomatonException):
     pass
 
 
-class DiagramException(Exception):
+class DiagramException(AutomatonException):
     """The diagram cannot be produced"""
 
     pass
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -17,6 +17,7 @@ from typing import (
     Literal,
     Set,
     Tuple,
+    TypeAlias,
     TypeVar,
     Union,
 )
@@ -30,6 +31,10 @@ except ImportError:
     _visual_imports = False
 else:
     _visual_imports = True
+finally:
+    # create a type for type checker
+    # irrespective of whether the imports were successful
+    GraphT: TypeAlias = "pgv.AGraph"
 
 
 LayoutMethod = Literal["neato", "dot", "twopi", "circo", "fdp", "nop"]
@@ -81,7 +86,7 @@ def create_graph(
     reverse_orientation: bool = False,
     fig_size: Union[Tuple[float, float], Tuple[float], None] = None,
     state_separation: float = 0.5,
-) -> pgv.AGraph:
+) -> GraphT:
     """Creates and returns a graph object
     Args:
         - horizontal (bool, optional): Direction of node layout. Defaults
@@ -118,7 +123,7 @@ def create_graph(
 
 
 def save_graph(
-    graph: pgv.AGraph,
+    graph: GraphT,
     path: Union[str, os.PathLike],
 ) -> None:
     """Write `graph` to file given by `path`. PNG, SVG, etc.
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -1,11 +1,36 @@
 #!/usr/bin/env python3
 """Miscellaneous utility functions and classes."""
 
+import os
+import pathlib
 from collections import defaultdict
 from itertools import count, tee, zip_longest
-from typing import Any, Callable, Dict, Generic, Iterable, List, Set, Tuple, TypeVar
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generic,
+    Iterable,
+    List,
+    Literal,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 from frozendict import frozendict
+
+# Optional imports for use with visual functionality
+try:
+    import pygraphviz as pgv
+except ImportError:
+    _visual_imports = False
+else:
+    _visual_imports = True
+
+
+LayoutMethod = Literal["neato", "dot", "twopi", "circo", "fdp", "nop"]
 
 
 def freeze_value(value: Any) -> Any:
@@ -39,6 +64,66 @@ def get_renaming_function(counter: count) -> Callable[[Any], int]:
     """
 
     return defaultdict(counter.__next__).__getitem__
+
+
+def create_graph(
+    horizontal: bool = True,
+    reverse_orientation: bool = False,
+    fig_size: Union[Tuple[float, float], Tuple[float], None] = None,
+    state_separation: float = 0.5,
+) -> pgv.AGraph:
+    """Creates and returns a graph object
+    Args:
+        - horizontal (bool, optional): Direction of node layout. Defaults
+            to True.
+        - reverse_orientation (bool, optional): Reverse direction of node
+            layout. Defaults to False.
+        - fig_size (tuple, optional): Figure size. Defaults to None.
+        - state_separation (float, optional): Node distance. Defaults to 0.5.
+    Returns:
+        AGraph with the given configuration.
+    """
+    if not _visual_imports:
+        raise ImportError(
+            "Missing visualization packages; "
+            "please install coloraide and pygraphviz."
+        )
+
+    # Defining the graph.
+    graph = pgv.AGraph(strict=False, directed=True)
+
+    if fig_size is not None:
+        graph.graph_attr.update(size=", ".join(map(str, fig_size)))
+
+    graph.graph_attr.update(ranksep=str(state_separation))
+
+    if horizontal:
+        rankdir = "RL" if reverse_orientation else "LR"
+    else:
+        rankdir = "BT" if reverse_orientation else "TB"
+
+    graph.graph_attr.update(rankdir=rankdir)
+
+    return graph
+
+
+def save_graph(
+    graph: pgv.AGraph,
+    path: Union[str, os.PathLike],
+) -> pgv.AGraph:
+    """Write `graph` to file given by `path`. PNG, SVG, etc.
+    Returns the same graph."""
+
+    save_path_final: pathlib.Path = pathlib.Path(path)
+
+    format = save_path_final.suffix.split(".")[1] if save_path_final.suffix else None
+
+    graph.draw(
+        path=save_path_final,
+        format=format,
+    )
+
+    return graph
 
 
 T = TypeVar("T")
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """Miscellaneous utility functions and classes."""
+from __future__ import annotations
 
 import os
 import pathlib
@@ -17,7 +18,6 @@ from typing import (
     Literal,
     Set,
     Tuple,
-    TypeAlias,
     TypeVar,
     Union,
 )
@@ -31,10 +31,6 @@ except ImportError:
     _visual_imports = False
 else:
     _visual_imports = True
-finally:
-    # create a type for type checker
-    # irrespective of whether the imports were successful
-    GraphT: TypeAlias = "pgv.AGraph"
 
 
 LayoutMethod = Literal["neato", "dot", "twopi", "circo", "fdp", "nop"]
@@ -86,7 +82,7 @@ def create_graph(
     reverse_orientation: bool = False,
     fig_size: Union[Tuple[float, float], Tuple[float], None] = None,
     state_separation: float = 0.5,
-) -> GraphT:
+) -> pgv.AGraph:
     """Creates and returns a graph object
     Args:
         - horizontal (bool, optional): Direction of node layout. Defaults
@@ -123,7 +119,7 @@ def create_graph(
 
 
 def save_graph(
-    graph: GraphT,
+    graph: pgv.AGraph,
     path: Union[str, os.PathLike],
 ) -> None:
     """Write `graph` to file given by `path`. PNG, SVG, etc.
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -3,6 +3,8 @@
 
 import os
 import pathlib
+import random
+import uuid
 from collections import defaultdict
 from itertools import count, tee, zip_longest
 from typing import (
@@ -66,6 +68,14 @@ def get_renaming_function(counter: count) -> Callable[[Any], int]:
     return defaultdict(counter.__next__).__getitem__
 
 
+def create_unique_random_id() -> str:
+    # To be able to set the random seed, took code from:
+    # https://nathanielknight.ca/articles/consistent_random_uuids_in_python.html
+    return str(
+        uuid.UUID(bytes=bytes(random.getrandbits(8) for _ in range(16)), version=4)
+    )
+
+
 def create_graph(
     horizontal: bool = True,
     reverse_orientation: bool = False,
@@ -110,7 +120,7 @@ def create_graph(
 def save_graph(
     graph: pgv.AGraph,
     path: Union[str, os.PathLike],
-) -> pgv.AGraph:
+) -> None:
     """Write `graph` to file given by `path`. PNG, SVG, etc.
     Returns the same graph."""
 
@@ -122,8 +132,6 @@ def save_graph(
         path=save_path_final,
         format=format,
     )
-
-    return graph
 
 
 T = TypeVar("T")
--- a/automata/fa/fa.py
+++ b/automata/fa/fa.py
@@ -20,7 +20,9 @@ try:
     import coloraide
     import pygraphviz as pgv
 except ImportError:
-    pass
+    _visual_imports = False
+else:
+    _visual_imports = True
 
 
 FAStateT = AutomatonStateT
@@ -72,6 +74,12 @@ class FA(Automaton, metaclass=abc.ABCMeta):
         Returns:
             AGraph corresponding to the given automaton.
         """
+
+        if not _visual_imports:
+            raise ImportError(
+                "Missing visualization packages; "
+                "please install coloraide and pygraphviz."
+            )
 
         # Defining the graph.
         graph = create_graph(
--- a/automata/fa/fa.py
+++ b/automata/fa/fa.py
@@ -4,13 +4,16 @@ from __future__ import annotations
 
 import abc
 import os
-import random
-import uuid
 from collections import defaultdict
 from typing import Any, Dict, Generator, List, Optional, Set, Tuple, Union
 
 from automata.base.automaton import Automaton, AutomatonStateT
-from automata.base.utils import LayoutMethod, create_graph, save_graph
+from automata.base.utils import (
+    LayoutMethod,
+    create_graph,
+    create_unique_random_id,
+    save_graph,
+)
 
 # Optional imports for use with visual functionality
 try:
@@ -78,13 +81,9 @@ class FA(Automaton, metaclass=abc.ABCMeta):
         font_size_str = str(font_size)
         arrow_size_str = str(arrow_size)
 
-        # we use a random uuid to make sure that the null node has a
-        # unique id to avoid colliding with other states.
-        # To be able to set the random seed, took code from:
-        # https://nathanielknight.ca/articles/consistent_random_uuids_in_python.html
-        null_node = str(
-            uuid.UUID(bytes=bytes(random.getrandbits(8) for _ in range(16)), version=4)
-        )
+        # create unique id to avoid colliding with other states
+        null_node = create_unique_random_id()
+
         graph.add_node(
             null_node,
             label="",
@@ -159,7 +158,7 @@ class FA(Automaton, metaclass=abc.ABCMeta):
 
         # Write diagram to file
         if path is not None:
-            graph = save_graph(graph, path)
+            save_graph(graph, path)
 
         return graph
 
--- a/automata/fa/fa.py
+++ b/automata/fa/fa.py
@@ -44,6 +44,10 @@ class FA(Automaton, metaclass=abc.ABCMeta):
         of the form (from_state, to_state, symbol)
         """
 
+        raise NotImplementedError(
+            f"iter_transitions is not implemented for {self.__class__}"
+        )
+
     def show_diagram(
         self,
         input_str: Optional[str] = None,
--- a/automata/fa/fa.py
+++ b/automata/fa/fa.py
@@ -4,57 +4,29 @@ from __future__ import annotations
 
 import abc
 import os
-import pathlib
 import random
 import uuid
 from collections import defaultdict
-from typing import Any, Dict, Generator, List, Literal, Optional, Set, Tuple, Union
+from typing import Any, Dict, Generator, List, Optional, Set, Tuple, Union
 
 from automata.base.automaton import Automaton, AutomatonStateT
+from automata.base.utils import LayoutMethod, create_graph, save_graph
 
 # Optional imports for use with visual functionality
 try:
     import coloraide
     import pygraphviz as pgv
 except ImportError:
-    _visual_imports = False
-else:
-    _visual_imports = True
+    pass
 
 
 FAStateT = AutomatonStateT
-LayoutMethod = Literal["neato", "dot", "twopi", "circo", "fdp", "nop"]
 
 
 class FA(Automaton, metaclass=abc.ABCMeta):
     """An abstract base class for finite automata."""
 
     __slots__ = tuple()
-
-    @staticmethod
-    def _get_state_name(state_data: FAStateT) -> str:
-        """
-        Get an string representation of a state. This is used for displaying and
-        uses `str` for any unsupported python data types.
-        """
-        if isinstance(state_data, str):
-            if state_data == "":
-                return "λ"
-
-            return state_data
-
-        elif isinstance(state_data, (frozenset, tuple)):
-            inner = ", ".join(FA._get_state_name(sub_data) for sub_data in state_data)
-            if isinstance(state_data, frozenset):
-                if state_data:
-                    return "{" + inner + "}"
-                else:
-                    return "∅"
-
-            elif isinstance(state_data, tuple):
-                return "(" + inner + ")"
-
-        return str(state_data)
 
     @staticmethod
     def _get_edge_name(symbol: str) -> str:
@@ -83,7 +55,7 @@ class FA(Automaton, metaclass=abc.ABCMeta):
         """
         Generates the graph associated with the given DFA.
         Args:
-            input_str (str, optional): String list of input symbols. Defaults to None.
+            - input_str (str, optional): String list of input symbols. Defaults to None.
             - path (str or os.PathLike, optional): Path to output file. If
               None, the output will not be saved.
             - horizontal (bool, optional): Direction of node layout. Defaults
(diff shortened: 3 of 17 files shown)


automata/pda/configuration.py
```diff
@@ -27,3 +28,16 @@ def __repr__(self) -> str:
         return "{}({!r}, {!r}, {!r})".format(
             self.__class__.__name__, self.state, self.remaining_input, self.stack
         )
+
+    def __eq__(self, other: Any) -> bool:
+        """Return True if two PDAConfiguration are equivalent"""
+        if not isinstance(other, PDAConfiguration):
+            raise NotImplemented
```
caleb531, line 35: @Vipul-Cariappa `return` should be used here instead of `raise`:

```suggestion
            return NotImplemented
```

Fore more details on why this is, see this StackOverflow post: [Why return NotImplemented instead of raising NotImplementedError](https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror)

docs/pda/class-pda.md
```diff
@@ -5,6 +5,20 @@
 The `PDA` class is an abstract base class from which all pushdown automata
 inherit. It can be found under `automata/pda/pda.py`.
 
+The `PDA` class has the following abstract methods:
+
+## PDA.show_diagram(self, input_str = None, path = None):
+
+Constructs and returns a pygraphviz `AGraph` corresponding to this PAD. If `input_str` is
```
caleb531, line 12: @Vipul-Cariappa Typo here. `PAD` should be `PDA`.

```suggestion
Constructs and returns a pygraphviz `AGraph` corresponding to this PDA. If `input_str` is
```

automata/pda/npda.py
```diff
@@ -111,6 +126,44 @@ def _get_next_configurations(
...
+                continue
+
+            for curr_step in steps[i]:
+                for next_step in self._get_next_configurations(curr_step):
+                    if next_step == path[-1]:
+                        path.append(curr_step)
+                        break
+                else:
+                    continue
+                break
```
caleb531, line 163: @Vipul-Cariappa What is the purpose of the `break` at the end of the `for` body here? I'm trying to understand the control flow logic you are trying to achieve.

caleb531:
@Vipul-Cariappa To answer your question, I would personally prefer that the stack operation show for each transition because in principle, you want the diagram to be a visual representation of the machine. And changes to the stack can drastically affect the machine's behavior for a PDA.

@eliotwrobson Do you have any thoughts on any of this?

Vipul-Cariappa:
@caleb531 I have updated the code to also display the stack operations. Results:

DPDA from docs with `input_str = "aabb"`:
image

NPDA from docs with `input_str = "abba"`:
image

automata/pda/pda.py
```diff
@@ -258,6 +262,58 @@ def show_diagram(
...
+        end_color: coloraide.Color,
+        font_size: str,
+        arrow_size: str,
+    ) -> pgv.AGraph:
+        """
+        Constructs stack for all the transitions in the `input_path` and
+        adds the constructed stacks into `graph`. Returns the same `graph`
+        """
+        from_node = input_path[0][0]
+        label = " | ".join(reversed(from_node.stack))  # type: ignore[arg-type]
```
Vipul-Cariappa, line 279: `mypy` gives an error here, so I have put an ignore comment. If you want me to remove the ignore comment that can be done. I will change the code to `reversed(from_node.stack.stack)` that will resolve the error from `mypy`.

The `mypy` error is:
```
automata/pda/pda.py:279: error: Argument 1 to "reversed" has incompatible type "PDAStack"; expected "SupportsLenAndGetItem[str]"  [arg-type]
automata/pda/pda.py:279: note: Following member(s) of "PDAStack" have conflicts:
automata/pda/pda.py:279: note:     Expected:
automata/pda/pda.py:279: note:         def __getitem__(self, int, /) -> str
automata/pda/pda.py:279: note:     Got:
automata/pda/pda.py:279: note:         def __getitem__(self, Union[int, slice], /) -> Union[str, Sequence[str]]
```

automata/pda/pda.py
```diff
@@ -258,6 +262,58 @@ def show_diagram(
...
+            fontsize=font_size,
+        )
+
+        interpolation = coloraide.Color.interpolate(
+            [start_color, end_color], space="srgb"
+        )
+
+        for i, (c, n) in enumerate(input_path, start=1):
+            from_node = n
+            label = " | ".join(reversed(from_node.stack))  # type: ignore[arg-type]
```
Vipul-Cariappa, line 294: The same thing also applies here.

automata/pda/configuration.py
Vipul-Cariappa: I have made the suggested changes.

docs/pda/class-pda.md
Vipul-Cariappa: I have fixed the typo.

automata/pda/npda.py
Vipul-Cariappa: I need the steps taken by the `NPDA` to construct the state machine diagram for the given input string. I am getting the `steps` from `read_input_stepwise`. But `read_input_stepwise` returns all the possible steps and not just the actual steps needed. So I am constructing the actual only necessary steps backwards. Starting from the final state, append it to the list `path` then loop through all the possible states from the previous to the final step and check which one of them will reach the final state, append that to `path`, and continue for all the steps.

In a `NPDA` there can be multiple possible next states for a given state. The first `break` is reached when a possible next state is the same as the actual next state needed. Due to this `continue` will not be executed and even the outer `for` loop will exit, by reaching the second `break`. Otherwise `continue` will run, checking the remaining possible states. Once we append the step, we need not check which other states will reach the next required state.

Simply put, once a step is appended, then we exit from both the `for` loops. If the inner `for` loop finishes without appending an element we will `continue` checking.

automata/pda/pda.py
caleb531: @Vipul-Cariappa Per the docs for `reversed()`, you should be able to eliminate the need for the ignore comment by adding a `__reversed__` method to the `PDAStack` class.

```python
def __reversed__(self) -> Iterator[str]:
    return reversed(self.stack)
```

This method should've really been present anyway, so I would ask that you please go ahead and add it to `PDAStack`.

automata/pda/configuration.py
```diff
@@ -27,3 +28,16 @@ def __repr__(self) -> str:
         return "{}({!r}, {!r}, {!r})".format(
             self.__class__.__name__, self.state, self.remaining_input, self.stack
         )
+
+    def __eq__(self, other: Any) -> bool:
+        """Return True if two PDAConfiguration are equivalent"""
+        if not isinstance(other, PDAConfiguration):
+            return NotImplemented
+
+        if (
```
eliotwrobson, line 37:
```suggestion
        return (
```
No need to return true and false in separate statements.

automata/pda/pda.py
```diff
@@ -2,18 +2,34 @@
...
+except ImportError:
+    _visual_imports = False
+else:
+    _visual_imports = True
+
 PDAStateT = AutomatonStateT
 PDATransitionsT = AutomatonTransitionsT
 PDAAcceptanceModeT = Literal["final_state", "empty_stack", "both"]
 
+LayoutMethod = Literal["neato", "dot", "twopi", "circo", "fdp", "nop"]
```
eliotwrobson, line 31: Could this type be put in a shared file somewhere, just so we don't have to have this duplicated?

automata/pda/pda.py
```diff
@@ -24,6 +40,280 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
     initial_stack_symbol: str
     acceptance_mode: PDAAcceptanceModeT
 
+    @staticmethod
+    def _get_state_name(state_data: PDAStateT) -> str:
```
eliotwrobson, line 44: Isn't this function identical to the other `_get_state_name` function? It would be great if that logic could be shared somehow, even if it means putting this function into a utility file or something.

automata/pda/pda.py
```diff
@@ -24,6 +40,280 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+        graph = pgv.AGraph(strict=False, directed=True)
+
+        if fig_size is not None:
+            graph.graph_attr.update(size=", ".join(map(str, fig_size)))
+
+        graph.graph_attr.update(ranksep=str(state_separation))
+        font_size_str = str(font_size)
+        arrow_size_str = str(arrow_size)
+
+        if horizontal:
```
eliotwrobson, line 158: Nit: Could this layout determination logic be put in a helper function? This seems like it's identical to what is being used in the FA class.

automata/pda/pda.py
```diff
@@ -24,6 +40,280 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+                from_node,
+                label=f" | {label}",
+                shape="record",
+                arrowsize=arrow_size,
+                fontsize=font_size,
+            )
+            graph.add_edge(
+                c,
+                n,
+                label=f"<<b>[<i>#{i}</i>]</b>>",
```
eliotwrobson, line 308: In the edges, it doesn't look like the closing `]` is being displayed. Could something weird be going on here?

eliotwrobson (changes requested):
Left some early comments. Overall looks good, but needs some test cases and I think some things can be simplified significantly. Thanks for the contribution @Vipul-Cariappa !

eliotwrobson:
> @Vipul-Cariappa To answer your question, I would personally prefer that the stack operation show for each transition because in principle, you want the diagram to be a visual representation of the machine. And changes to the stack can drastically affect the machine's behavior for a PDA.

I'm in agreement @caleb531 , it's nice to be able to generate the states of the stack along with the visual representation of the machine. @Vipul-Cariappa, it would be awesome to have a flag enabling displaying only the machine or only the stack (both by default).

Vipul-Cariappa:
> it would be awesome to have a flag enabling displaying only the machine or only the stack (both by default).

@eliotwrobson, I have implemented the feature you asked for.

---

I have refactored the code to remove duplications, to the best of my abilities, according to what I felt makes sense.

Please have a look at it, and let me know if any other changes are required.

Two things left to do:
- Write test cases
- Update the md doc file for `with_machine` and `with_stack` parameters

automata/pda/npda.py
caleb531: @Vipul-Cariappa Fair enough! I am curious if this logic can be rewritten to be cleaner, but I will wait until after tests are written and this PR is merged. 🙂

Resolving this thread.

automata/pda/dpda.py
```diff
@@ -151,6 +162,31 @@ def _get_next_configuration(self, old_config: PDAConfiguration) -> PDAConfigurat
...
+        self, input_str: str
+    ) -> Tuple[List[Tuple[PDAConfiguration, PDAConfiguration]], bool]:
+        """
+        Calculate the path taken by input.
+
+        Args:
+            input_str (str): The input string to run on the DPDA.
+
+        Returns:
+            tuple[list[tuple[PDAConfiguration, PDAConfiguration], bool]]: A list
```
caleb531, line 175: @Vipul-Cariappa It's a small thing, but could we please match the casing of the actual type here? Also, I think the brackets are mismatched compared to the actual annotation.

```suggestion
            Tuple[List[Tuple[PDAConfiguration, PDAConfiguration]], bool]: A list
```

automata/pda/npda.py
```diff
@@ -111,6 +125,44 @@ def _get_next_configurations(
...
+        self, input_str: str
+    ) -> Tuple[List[Tuple[PDAConfiguration, PDAConfiguration]], bool]:
+        """
+        Calculate the path taken by input.
+
+        Args:
+            input_str (str): The input string to run on the NPDA.
+
+        Returns:
+            tuple[list[tuple[PDAConfiguration, PDAConfiguration], bool]]: A list
```
caleb531, line 138: @Vipul-Cariappa Same request as the other comment, if you could, please:

```suggestion
            Tuple[List[Tuple[PDAConfiguration, PDAConfiguration]], bool]: A list
```

Vipul-Cariappa:
I have added test cases. Again, I adapted test cases from `dfa` and `nfa`. Please review it and let me know if any more changes are required.

automata/base/utils.py
```diff
@@ -41,6 +66,66 @@ def get_renaming_function(counter: count) -> Callable[[Any], int]:
...
+    save_path_final: pathlib.Path = pathlib.Path(path)
+
+    format = save_path_final.suffix.split(".")[1] if save_path_final.suffix else None
+
+    graph.draw(
+        path=save_path_final,
+        format=format,
+    )
+
+    return graph
```
eliotwrobson, line 126: Why is the graph being returned here? I don't think it's being modified by this function.

automata/pda/pda.py
```diff
@@ -2,17 +2,40 @@
...
+from automata.base.utils import LayoutMethod, create_graph, save_graph
 from automata.pda.configuration import PDAConfiguration
 from automata.pda.stack import PDAStack
 
+# Optional imports for use with visual functionality
+try:
+    import coloraide
+    import pygraphviz as pgv
+except ImportError:
+    pass
```
eliotwrobson, line 33: Should use the same error message as in other files if the imports are not present.

automata/fa/fa.py
```diff
@@ -4,58 +4,30 @@
...
 
 # Optional imports for use with visual functionality
 try:
     import coloraide
     import pygraphviz as pgv
 except ImportError:
-    _visual_imports = False
-else:
-    _visual_imports = True
+    pass
```
eliotwrobson, line 20: Why use a pass here? If the imports are still being used somewhere, you should still set this boolean variable and throw an exception if a visualization function gets called and these fail. It's ok to remove the pygraphviz import if that's not being used here however.

automata/pda/npda.py
```diff
@@ -111,6 +125,44 @@ def _get_next_configurations(
...
+
+        """
+
+        steps = list(self.read_input_stepwise(input_str))
+
+        path: List[PDAConfiguration] = [steps.pop().pop()]
+
+        accepted = path[0] in self.final_states
+
+        for i in range(len(steps) - 1, -1, -1):
```
eliotwrobson, line 150: I think you can just iterate through `steps` backwards instead of indexing, so just do `reversed(steps)`.

automata/pda/pda.py
```diff
@@ -24,6 +47,313 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+        """
+        Iterate over all transitions in the automaton.
+        Each transition is a tuple of the form
+        (from_state, to_state, (input_symbol, stack_top_symbol, stack_push_symbols))
+        """
+
+    def show_diagram(
+        self,
+        input_str: Optional[str] = None,
+        with_machine: bool = True,
```
eliotwrobson, line 96: Should add an error if both `with_machine` and `with_stack` are both false (since this almost certainly means a mistake in client code).

automata/pda/pda.py
```diff
@@ -24,6 +47,313 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+            - reverse_orientation (bool, optional): Reverse direction of node
+              layout. Defaults to False.
+            - fig_size (tuple, optional): Figure size. Defaults to None.
+            - font_size (float, optional): Font size. Defaults to 14.0.
+            - arrow_size (float, optional): Arrow head size. Defaults to 0.85.
+            - state_separation (float, optional): Node distance. Defaults to 0.5.
+        Returns:
+            AGraph corresponding to the given automaton.
+        """
+
```
eliotwrobson, line 129: See previous comment, should add an error message for missing imports.

automata/pda/pda.py
```diff
@@ -24,6 +47,313 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+        self,
+        graph: pgv.AGraph,
+        arrow_size: str,
+        font_size: str,
+    ) -> pgv.AGraph:
+        """
+        Add all the states of the PDA
+        """
+
+        # we use a random uuid to make sure that the null node has a
```
eliotwrobson, line 328: Could this UUID generation be moved to a helper function? this exact line with comment appears in a few places.

automata/pda/pda.py
```diff
@@ -24,6 +47,313 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+
+    @abc.abstractmethod
+    def iter_transitions(
+        self,
+    ) -> Generator[Tuple[PDAStateT, PDAStateT, Tuple[str, str, str]], None, None]:
+        """
+        Iterate over all transitions in the automaton.
+        Each transition is a tuple of the form
+        (from_state, to_state, (input_symbol, stack_top_symbol, stack_push_symbols))
+        """
```
eliotwrobson, line 91: I think there should be a not implemented error here?

eliotwrobson (changes requested):
New changes are looking good, in particular the tests seem fairly comprehensive! Left comments in a few places, I think some logic can be simplified significantly and there were a few things I had questions about.

automata/pda/pda.py
Vipul-Cariappa: The closing `]` is not displayed even in DFA.

image


Not sure but maybe its a bug with `pygraphviz`?

automata/base/utils.py
Vipul-Cariappa: Yes. Makes sense. I will change it so that it doesn't return anything.

automata/fa/fa.py
Vipul-Cariappa: The `pygraphviz` import is required for the type hints. `pygraphvix.AGraph` is used in this file/module. But any call to `pygraphviz` first goes through `create_graph` in `automata.base.utils` which raises an appropriate error message.
Therefore I thought of removing it here. And from the other file.

If you would want me to add it here too, I can do that. Let me know.

---

Found an interesting thing.

Even though I have `try` and `except` for the optional feature imports. I get a `NameError` if `pygraphviz` is not installed.

```bash
>>> from automata.fa.dfa import DFA
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/vipul/workspace/python/automata/automata/fa/dfa.py", line 33, in <module>
    import automata.fa.fa as fa
  File "/home/vipul/workspace/python/automata/automata/fa/fa.py", line 12, in <module>
    from automata.base.automaton import Automaton, AutomatonStateT
  File "/home/vipul/workspace/python/automata/automata/base/automaton.py", line 12, in <module>
    from automata.base.utils import freeze_value
  File "/home/vipul/workspace/python/automata/automata/base/utils.py", line 74, in <module>
    ) -> pgv.AGraph:
         ^^^
NameError: name 'pgv' is not defined
```

> Error from `utils.py`

I am thing to fix the above error by doing the following:

```python
try:
    from pygraphviz import AGraph
except ImportError:
    AGraph = None
```

Any thoughts?

---

This is should have been caught from the automated test runner. I would recommend to run the test cases with the optional dependencies and without it. To catch any such errors with may raise due to the absence of optional dependencies.

automata/pda/pda.py
Vipul-Cariappa: ok. Will do that

automata/pda/pda.py
Vipul-Cariappa: Yes, it makes sense. I will update.

automata/pda/pda.py
Vipul-Cariappa: The first line in this function calls `create_graph` which takes care of raising the correct error message.

If you want me to add it to this function anyway, I can do it. Let me know.

Vipul-Cariappa:
I have made some of the changes @eliotwrobson suggested. Also, left some clarifications and questions. 
Please have a look at it.

automata/pda/pda.py
eliotwrobson: Oh hmmm that's a little strange. I'll play with the current version locally to see what's up

automata/fa/fa.py
eliotwrobson: Ahh I see. Although the solution to this issue usually comes up in the context of circular imports, I think we can do the following: https://stackoverflow.com/a/39757388/2923069

> This is should have been caught from the automated test runner. I would recommend to run the test cases with the optional dependencies and without it. To catch any such errors with may raise due to the absence of optional dependencies.

This is unfortunately a little involved. To do this, we have to set up tox to run an additional test suite that's customized to check for desired behavior if optional dependencies are not included. I'd absolutely welcome an issue / PR doing this, as it's also a prerequisite for #159, but may be a little beyond the scope of this PR right now.

automata/pda/pda.py
eliotwrobson: It looks like `create_graph` only raises an error message if `pygraphviz` is missing, but this function uses `coloraide` as well, so I think we need the additional check.

Vipul-Cariappa:
I guess all change requests are implemented. Please have a look. Let me know if I missed any.

automata/base/utils.py
```diff
@@ -1,12 +1,44 @@
...
+try:
+    import pygraphviz as pgv
+except ImportError:
+    _visual_imports = False
+else:
+    _visual_imports = True
+finally:
+    # create a type for type checker
+    # irrespective of whether the imports were successful
+    GraphT: TypeAlias = "pgv.AGraph"
```
eliotwrobson, line 37: I don't think that there is a need to use a string annotation here. I believe you can delete this block and add `from __future__ import annotations` at the top (as this will not read annotations unless code is being explicitly typechecked).

automata/pda/pda.py
```diff
@@ -24,6 +52,325 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+
+        return graph
+
+    def _add_transitions_diagram(
+        self,
+        graph: pgv.AGraph,
+        is_edge_drawn: EdgeDrawnDictT,
+        arrow_size: str,
+        font_size: str,
+    ) -> pgv.AGraph:
```
eliotwrobson, line 313: I think these annotations might cause issues without `from __future__ import annotations` at the top of this file.

automata/pda/pda.py
```diff
@@ -24,6 +52,325 @@ class PDA(Automaton, metaclass=abc.ABCMeta):
...
+            tooltip="->" + initial_node,
+            arrowsize=arrow_size,
+        )
+
+        nonfinal_states = map(self._get_state_name, self.states - self.final_states)
+        final_states = map(self._get_state_name, self.final_states)
+        graph.add_nodes_from(nonfinal_states, shape="circle", fontsize=font_size)
+        graph.add_nodes_from(final_states, shape="doublecircle", fontsize=font_size)
+
+        return graph
```
eliotwrobson, line 372: Aren't all of these changes made to the object in-place? If so, there's no need to return anything. Same goes for all of the other helper functions that are returning the graph.

automata/pda/npda.py
```diff
@@ -111,6 +125,44 @@ def _get_next_configurations(
...
+        for i in reversed(steps):
+            if len(i) == 1:
+                path.append(i.pop())
+                continue
+
+            for curr_step in i:
+                for next_step in self._get_next_configurations(curr_step):
+                    if next_step == path[-1]:
+                        path.append(curr_step)
+                        break
```
eliotwrobson, line 159: A little bit of a nitpick, but I think now there are tests for this and it has already been simplified somewhat, it would be good to see if we could simplify these cases a little bit before merging.

automata/pda/npda.py
```diff
@@ -111,6 +125,44 @@ def _get_next_configurations(
...
+
+        """
+
+        steps = list(self.read_input_stepwise(input_str))
+
+        path: List[PDAConfiguration] = [steps.pop().pop()]
+
+        accepted = path[0] in self.final_states
+
+        for i in reversed(steps):
```
eliotwrobson, line 150:
```suggestion
        for step in reversed(steps):
```
Slightly clearer variable name.

eliotwrobson (changes requested):
Don't want to approve quite yet (as I'd like a little more time to review everything, and requested a few changes), but this is getting very close to being ready to merge!

automata/pda/npda.py
Vipul-Cariappa: I believe it is simple enough. Stared at the code for 2 mins and could not think of anything. Do you have anything in mind to simplify this.

eliotwrobson: So the `_get_next_configurations(curr_step)` function returns a set, and you're just checking whether `path[-1]` is contained in there. So I think you can just do `if path[-1] in self._get_next_configurations(curr_step)` instead of the innermost loop, and then do `path.append(curr_step)` inside the new if block.

Vipul-Cariappa: @eliotwrobson Sorry, for the late response. I have implemented your suggestions. Please have a look. The code is way easier to read with this modification you specified. ☺️

caleb531: @eliotwrobson Please take a look at @Vipul-Cariappa's changes, and if you are satisfied, feel free to resolve this conversation thread.

eliotwrobson: New code looks great! Will resolve and thanks for your patience @Vipul-Cariappa

eliotwrobson (approved):
I want to double check whatever is going on with the closing bracket in the unresolved discussion, but the code looks good!

automata/pda/pda.py
eliotwrobson: Maybe this is a graphviz thing, but it looks fine on my machine:
image

Will go ahead and resolve.

caleb531:
@eliotwrobson I won't have time to review it today, but I can probably look tomorrow or Monday. In the meantime, could you please add some tests for the `NotImplemented` / `NotImplementedError` cases? That will bring the coverage back up, and there should be some existing tests you can use as reference.

tests/test_npda.py
```diff
@@ -390,3 +392,149 @@ def test_accepts_input_true(self) -> None:
...
+                    }
+                ),
+                "q2",
+            ),
+        }
+        seen_transitions = {
+            (edge[0], frozenset(edge.attr["label"].split("\n")), edge[1])
+            for edge in graph.edges()
+        }
+        self.assertTrue(expected_transitions.issubset(seen_transitions))
```
eliotwrobson, line 450: @Vipul-Cariappa not sure why, but this check in the test cases fails locally for me. Since this is an npda, I suspect there is a mistake in the algorithm or test somewhere that's causing non-deterministic failures.

Vipul-Cariappa: @eliotwrobson All the tests pass on my local machine, also on GitHub workflow. Could you please attach the traceback for more context? It would be easier to debug. It will be impossible to debug without it.

eliotwrobson: Here's a screenshot with the error message:
image

The set that's being printed here is `seen_transitions - expected_transitions`, so the transitions that are being produced are not a subset of those that are expected. Not quite sure why this is. Here is console output of me printing the two sets in question:

image

eliotwrobson: @Vipul-Cariappa Figured this out and fixed in this commit: https://github.com/caleb531/automata/pull/177/commits/b52775ee5423bd147fa76e568dfca7df7b460619

I think graphviz was adding some extra characters causing the test to fail, but this depends on how this is implemented under the hood. Either way, going to resolve this discussion!

caleb531 (changes requested):
@eliotwrobson @Vipul-Cariappa This PR looks very good! I only have one requested change (well, technically two, I suppose) related to the new `DiagramException` class. But if one of you could please make those couple changes quick, I'll be ready to give this the final sign-off!

Thanks also to both of you for all your hard work on this PR so far.

automata/base/exceptions.py
```diff
@@ -87,3 +87,9 @@ class InfiniteLanguageException(AutomatonException):
     """The operation cannot be performed because the language is infinite"""
 
     pass
+
+
+class DiagramException(Exception):
```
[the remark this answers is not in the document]
eliotwrobson: @caleb531 done here: https://github.com/caleb531/automata/pull/177/commits/510e4716457c918a2395568e445a26fa4440746d

caleb531 (approved):
@eliotwrobson Thanks for that! Everything looks good to me now. Approving and merging...

PR #180, opened 2023-10-25
Generate documentation site using Mkdocs

lewiuberg:
Starting point for using mkdocs to generate documentation site.


automata/base/automaton.py
```diff
@@ -37,72 +40,188 @@ def __init__(self, **kwargs: Any) -> None:
...
+            This method does not return anything as it always raises.
+
+        Raises
+        ------
+        AttributeError
+            If this method is called.
+        """
         raise AttributeError(f"This {type(self).__name__} is immutable")
 
     def __delattr__(self, name: str) -> None:
```
eliotwrobson, line 84: @caleb531 any thoughts on whether these overrides should appear in the public API? I'm not sure if there's a way to make them hidden in the docs or something.

eliotwrobson (approved):
Looks great so far, thanks for your work @lewiuberg ! One small question but generally ok with being generous merging to this feature branch.

automata/base/automaton.py
caleb531: @eliotwrobson I think they would just add noise and therefore shouldn't be documented, but my assumption was that mkdocs would generate the docs site based on the Markdown documentation we already have, not from the docstrings and class/method signatures (the latter would be devoid of all the examples and explanation we've written out).

eliotwrobson: So this PR adds a plugin that generates API documentation for individual methods from docstrings. So any discussion from the current documentation can be copied to the docstrings. I wonder if there's a way to get mkdocstrings ignore certain functions.

caleb531: @eliotwrobson Hm, I'm not sure I'm in favor of this way of generating documentation, then. It would add substantial weight to the Python source files. This would then make the Python files more difficult to skim through, and we'd lose Markdown syntax highlighting (which would increase the possibility of syntax errors). I can also foresee some current sections of the docs not translating over as cleanly as other sections.

My personal preference would be to wrap the existing documentation that we've written (i.e. every Markdown file in the `docs/` directory) inside of a GitHub Pages site. My apologies if I've misunderstood mkdocs's purpose the whole time.

eliotwrobson: @caleb531 I think that's a valid concern. The issue is that PyOpenSci requires docstrings (or at least, strongly suggests them, see [here](https://www.pyopensci.org/python-package-guide/documentation/write-user-documentation/document-your-code-api-docstrings.html#what-is-a-docstring-and-how-does-it-relate-to-documentation)), and I think these would be a generally good thing to add to the library. Then, maintaining separate API documentation like we have been seems like a substantial maintenance overhead, as we'll have to maintain the docstrings and API instead of having one generated from the other. This is my mistake however, I didn't do a good job of clearly communicating the docstring requirement. What are your thoughts on this? I'm actually a bit torn as well, but I figured that if we have to make docstrings anyway it's best to reduce duplication wherever possible.

I think your overall understanding is correct though, mkdocs is just generating the GitHub Pages site for documentation, but the particular extension in this PR is what is generating the API documentation from the docstrings.

lewiuberg: @caleb531 Just to be clear, its mkdocstrings that generates documentation using docstrings. I made an example using it (API) after @eliotwrobson suggestion to use it. I always write NumPy docstrings, so it makes sense to me. The first PR I sent I had another example (API Documentation) where I use the documentation that already were in the repo.
If you switch to the branch `docs_rework`, install the requirements and run `mkdocs serve` you can se both examples in action.

eliotwrobson: @caleb531 also, regarding syntax / correctness of examples used in documentation, using docstrings I think we can just use [doctest](https://docs.python.org/3/library/doctest.html) for this, and then it will carry over to the generated docs as well.

caleb531: @eliotwrobson Okay, then. I'd be open for trying this. Perhaps since migrating the entire documentation to docstrings will be a sizable amount of work, perhaps we should start with a few select classes/methods, and see how that looks under this mkdocstrings-based website.

cc @lewiuberg

eliotwrobson: I think that's a good plan. Starting with this PR, we should be able to do this gradual migration on the feature branch.

eliotwrobson: Since we seem to have reached a consensus here and this is just a feature branch PR, I'm going to go ahead and merge. We can migrate other files as needed in later PRs on this same branch.

caleb531: @eliotwrobson Yep, sounds good. Please proceed!

PR #181, opened 2023-10-29
Fix bug: Replace * with *? in QuantifierToken regex pattern

christopher-hampson:
Greedy matching for the QuantifierToken appears to cause issues for patterns with more than one quantifier.

### DIFF
--- a/automata/regex/parser.py
+++ b/automata/regex/parser.py
@@ -551,7 +551,7 @@ def get_regex_lexer(
     lexer.register_token(KleeneStarToken.from_match, r"\*")
     lexer.register_token(KleenePlusToken.from_match, r"\+")
     lexer.register_token(OptionToken.from_match, r"\?")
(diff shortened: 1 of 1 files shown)


eliotwrobson:
@christopher-hampson thank you for your contribution! Do you happen to have any examples where the old code had issues fixed by the new code? It would be nice to have a test case for the behavior specifically for the behavior this is addressing.

caleb531 (changes requested):
@christopher-hampson Yes, I would agree with @eliotwrobson that an additional test case to verify these changes would be preferable, if you could please provide one. This would demonstrate, in a TDD fashion, that the hypothetical omission of these lexer changes would break this new test.

christopher-hampson:
Thanks @eliotwrobson @caleb531, I've added a couple additional tests to test_quantifier which seemed the most appropriate place for this but apologies if not.

eliotwrobson:
@christopher-hampson it looks like the tests are failing

eliotwrobson (approved):

Issue #161, opened 2023-07-18
Valmari's Algorithm for Minimizing Partial DFAs 


eliotwrobson:
https://arxiv.org/abs/0802.2826

This paper is a more abstract description of Valmari's algorithm for minimizing partial DFAs. It should be easier to read and implement than the paper with the C++ code. Once we get a first iteration of partial DFA support, it would be awesome if we could add this algorithm as the default for minimizing partial DFAs (should be _much_ faster over large alphabets).

_Originally posted by @eliotwrobson in https://github.com/caleb531/automata/issues/126#issuecomment-1595611839_

sracha4355:
Hey I could give it a shot. Are you guys under any time constraint to get this done? Asking because I will be busy with school.

eliotwrobson:
@sracha4355 that would be awesome! No time constraint, and feel free to put up a draft PR if you ever want any feedback.

sracha4355:
Any tips on understanding the codebase, I am not an active user. I was just looking for cool projects to contribute to!

eliotwrobson:
This is actually a fairly good task for someone who is not familiar, as it only involves implementing a single algorithm. The main thing is to take a look at the `_minify` function in the DFA class, as the new algorithm will be replacing the function there. The other main thing is to understand is how DFAs are represented in the library, and for that I think it should be sufficient to take a look at some of the examples in the documentation: https://github.com/caleb531/automata/blob/main/docs/fa/examples.md

EduardoGoulart1:
> @EduardoGoulart1 Hey, I was planning on tackling this issue once school ends, which is in 2 weeks. But if you would like to work on this, go ahead! If you want to collaborate let me know.

@sracha4355 Could you please update us here when you start to work on it? I can wait until the turn of the year. But it would be good to have it merged rather soonish

sracha4355:
@EduardoGoulart1 you can go ahead and work on it!

eliotwrobson:
Don't worry too much about doing overlapping work. I think this algorithm is pretty tricky to implement correctly and might even take a few attempts to get right.

EduardoGoulart1:
Hm ok. TBH I'm quite surprised that the speedup is so small for large alphabets...

PR #183, opened 2023-11-04
Fix issue with minimization of partial DFAs

### LINKED ISSUE #182: Minimization of DFA from finite language makes the language infinite
While using the `DFA.from_finite_language()` method to create a basic DFA from a finite language and then applying dfa.minify(), the resulting DFA incorrectly introduces a loop that allows an infinite number of accepted words. This behavior deviates from the expected result where the minimized DFA should recognize the same finite language as the unminimized DFA.

**Steps to Reproduce:**

1. Create a DFA from a finite language using DFA.from_finite_language() with the language {'ab', 'abcb'} and input symbols {'a', 'b', 'c'}.
1. Print the maximum word length and the DFA representation.
1. Apply the minify() method on the DFA.
1. Print the maximum word length and the minimized DFA representation.

**Code**

```python
dfa = DFA.from_finite_language(
    language={'ab', 'abcb'},
    input_symbols={'a', 'b', 'c'}
)
print("unminified maximum_word_length:", dfa.maximum_word_length())
print("unminified:", dfa)
dfa = dfa.minify()
print("minified maximum_word_length:", dfa.maximum_word_length())
print("minified:", dfa)
```

**Output**
```shell
unminified maximum_word_length: 4
unminified: DFA(states={'', 'ab', 'abc', 'abcb', 'a'}, input_symbols={'a', 'c', 'b'}, transitions={'': {'a': 'a'}, 'a': {'b': 'ab'}, 'ab': {'c': 'abc'}, 'abc': {'b': 'abcb'}, 'abcb': {}}, initial_state='', final_states={'ab', 'abcb'}, allow_partial=True)
minified maximum_word_length: None
minified: DFA(states={0, 1, 2}, input_symbols={'a', 'c', 'b'}, transitions={0: {'a': 2}, 1: {'c': 2}, 2: {'b': 1}}, initial_state=0, final_states={1}, allow_partial=True)
DFA(states={0, 1, 2, 3}, input_symbols={'a', 'c', '<E>', 'b'}, transitions={0: {'a': 2}, 1: {'c': 2, '<E>': 3}, 2: {'b': 1}, 3: {}}, initial_state=0, final_states={3}, allow_partial=True)
```

**DFA**

**Minified DFA**

**Environment**:

- Python version: 3.9
- automata-lib version: 8.1.0
- Operating System: Debian

The problem seems to be related to the minimization algorithm, potentially in the way it handles transitions and final states during state reduction, but I may have missed something, thanks for your help !

eliotwrobson:
See title, resolves #182. I think this was caused by issues related to running the minimization algorithm on partial DFAs. This changes the API slightly, but should resolve those issues. Still a draft because docs changes are needed. Slight decrease in efficiency but I think this isn't going to affect very many people. We should be able to change things back somewhat once #161 gets merged and we have an algorithm specifically written for partial DFAs.

cc @martinec

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -454,7 +454,7 @@ class DFA(fa.FA):
 
         else:
             # Compute reachable states and final states
-            transitions = self.transitions
+            transitions = cast(DFATransitionsT, self.transitions)
 
         bfs_states = self.__class__._bfs_states(
             self.initial_state, lambda state: iter(transitions[state].items())
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -224,7 +224,7 @@ class DFA(fa.FA):
         """Returns the cardinality of the language represented by the DFA."""
         return self.cardinality()
 
-    def to_partial(self, *, retain_names: bool = False, minify: bool = True) -> Self:
+    def to_partial(self, *, retain_names: bool = False) -> Self:
         """
         Turns a DFA (complete or not) into a partial DFA.
         Removes dead states and trap states (except the initial state)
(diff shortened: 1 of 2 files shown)


eliotwrobson:
Closing to remake later. Nothing wrong with the approach here, but I think that there's a way of doing this that handles edge cases more cleanly.

PR #184, opened 2023-11-07
fix: Minify for partial DFAs

### LINKED ISSUE #182: Minimization of DFA from finite language makes the language infinite

eliotwrobson:
Resolves #182. Modifies the algorithm to deal with partial DFAs correctly. Somewhat inefficient, but the best we can do until Valmari's algorithm gets implemented (which is a separate open issue).

cc @martinec. If you have any more potentially interesting test cases, would be good to add to make sure the algorithm is working correctly (this one is unfortunately a bit tricky to get right).

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -514,7 +514,6 @@ class DFA(fa.FA):
                 for symbol in input_symbols:
                     end_state = path.get(symbol)
                     if end_state is not None:
-                        # for symbol, end_state in path.items():
                         symbol_dict = transition_back_map[symbol]
                         # If statement here needed to ignore certain transitions
                         # for non-reachable states
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -599,7 +599,7 @@ class DFA(fa.FA):
             new_transitions[name] = {
                 letter: back_map[inner_transition_dict_old[letter]]
                 for letter in inner_transition_dict_old.keys()
-                if inner_transition_dict_old[letter] in reachable_states
+                if inner_transition_dict_old[letter] in back_map.keys()
             }
 
         allow_partial = any(
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -499,6 +499,8 @@ class DFA(fa.FA):
         If the input DFA is partial, then the result is also a partial DFA
         """
 
+        reachable_states = set(reachable_states)
+
         # Per input-symbol backmap (tgt -> origin states)
         transition_back_map: Dict[str, Dict[DFAStateT, List[DFAStateT]]] = {
             symbol: {end_state: [] for end_state in reachable_states}
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -499,28 +499,45 @@ class DFA(fa.FA):
         If the input DFA is partial, then the result is also a partial DFA
         """
 
-        # First, assemble backmap and equivalence class data structure
+        # Per input-symbol backmap (tgt -> origin states)
+        transition_back_map: Dict[str, Dict[DFAStateT, List[DFAStateT]]] = {
+            symbol: {end_state: [] for end_state in reachable_states}
+            for symbol in input_symbols
+        }
+
+        trap_state = None
+
+        for start_state, path in transitions.items():
+            if start_state in reachable_states:
+                for symbol in input_symbols:
+                    end_state = path.get(symbol)
+                    if end_state is not None:
+                        # for symbol, end_state in path.items():
+                        symbol_dict = transition_back_map[symbol]
+                        # If statement here needed to ignore certain transitions
+                        # for non-reachable states
+                        if end_state in symbol_dict:
+                            symbol_dict[end_state].append(start_state)
+                    else:
+                        # Add trap state if needed
+                        if trap_state is None:
+                            trap_state = next(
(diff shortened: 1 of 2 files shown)


automata/fa/dfa.py
```diff
@@ -499,28 +499,47 @@ def _minify(
...
-                for symbol, end_state in path.items():
-                    symbol_dict = transition_back_map[symbol]
-                    # If statement here needed to ignore certain transitions
-                    # when minifying a partial DFA.
-                    if end_state in symbol_dict:
-                        symbol_dict[end_state].append(start_state)
+                for symbol in input_symbols:
+                    end_state = path.get(symbol)
+                    if end_state is not None:
+                        # for symbol, end_state in path.items():
```
EduardoGoulart1, line 517: [minor] remove this comment

tests/test_dfa.py
```diff
@@ -1280,6 +1280,35 @@ def test_minify_partial_dfa(self) -> None:
         self.assertEqual(len(minified_partial_dfa.states), 4)
         self.assertEqual(minified_partial_dfa, partial_dfa_extra_state)
 
+    def test_minify_partial_dfa_correctness(self) -> None:
+        """Test correctness of minifying partial DFAs"""
```
EduardoGoulart1, line 1284: For further context, I would add a link to the GitHub issue

EduardoGoulart1 (approved):
I am currently also struggling with this bug. This LGTM as a temporary fix.

eliotwrobson:
@caleb531 since this is a bug people are actively running into and the volume of changes (outside of the test cases) is small, I'm going to try and merge this sometime today.

automata/fa/dfa.py
eliotwrobson: Removed here: https://github.com/caleb531/automata/pull/184/commits/ecd87c4e0350cf01e2d721ba5c61b7e9cd2a39f1

tests/test_dfa.py
eliotwrobson: Added comment here: https://github.com/caleb531/automata/pull/184/commits/ecd87c4e0350cf01e2d721ba5c61b7e9cd2a39f1

caleb531 (approved):

caleb531:
@eliotwrobson Sure, I have no problem with that—the PR looks fine to me. And if you want to go ahead and merge to `main as well, please feel free. Just please be sure to run through all steps in the [Release Checklist](https://github.com/caleb531/automata/blob/main/.github/RELEASE_CHECKLIST.md) (including the creation of the GitHub Release with recent features like the PDA `show_diagram` method, etc.). I just want to make sure that we have everything captured.

eliotwrobson:
@caleb531 going to go ahead and merge since you approved.

I don't quite have time to run through the whole release checklist right now, since we added a couple of new features in addition to this bug fix. I'll try and find time in the next couple of days to push out a minor version. I think the main thing is that people who run into this bug have a fixed version available, and just having that on the develop branch should at least help some.

caleb531:
@eliotwrobson Oh okay, sure. If you think having this fix on `develop` is sufficient, then I'd probably not rather rush a release if we truly don't need to.

PR #191, opened 2023-12-06
Move CONTRIBUTING.md and add badges

eliotwrobson:
See title, more changes in prep for the upcoming review. Normally wouldn't care too much about this file placement, but this is specifically what is asked fro in their guidelines: https://www.pyopensci.org/python-package-guide/documentation/repository-files/contributing-file.html

Also added badges that are missing from the reviewer checklist: https://www.pyopensci.org/software-peer-review/how-to/reviewer-guide.html#peer-review-template

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,10 @@
 # Contributing to Automata
 
+## Code of Conduct
+
+When interacting with other users and maintainers, please be sure to abide by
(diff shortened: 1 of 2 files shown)


caleb531 (approved):
Yep that's fine! I have no problems with moving CONTRIBUTING.md to the project root, and think it's actually more discoverable that way, anyway.

PR #194, opened 2023-12-22
Correct typo in allow_mutable_automata documentation

anshravalll:
Changed 'enable_mutable_automata' global configuration to 'allow_mutable_automata' as mentioned in base/config.py

PR #196, opened 2023-12-30
Adding back missing docs

eliotwrobson:
Added back missing docs page from move to mkdocs. Uses change from #194

### DIFF
--- a/docs/characteristics.md
+++ b/docs/characteristics.md
@@ -1,0 +1,77 @@
+# Automaton Characteristics
+
+## Automaton instances are immutable
+
+All Automaton instances are fully immutable to protect against common pitfalls,
+such as mutating an automaton to an invalid state after it's already been
+validated.
+
+This means that if you wish to make a change to an automaton instance, you must
+retrieve its attributes as a dictionary (using the `input_parameters` property),
+make your desired change, then pass those parameters to the relevant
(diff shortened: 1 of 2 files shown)


caleb531 (approved):
This looks good! But quick question, and something I've been wondering:

I suppose I was under the impression that we'd be converting _all_ of the Markdown pages to docstrings within the code. But it seems like we'll be keeping both? Can you please clarify where Markdown is being replaced by docstrings and where it isn't?

eliotwrobson:
@caleb531 to your question, I think it makes sense to at least keep all of the API documentation in docstrings and have that generated by mkdocs. Putting everything in a docstring doesn't make sense for some of the documentation, especially in the case of long examples or extended discussions, such as the automaton characteristics here. However, I think this is the only case where there was some extended discussion in the old API documentation that didn't really fit into a docstring.

In this case, I think the change is more organizational than anything else (as the new docs page describes general behavior rather than specific methods), and it's questionable whether this discussion really should have ever been included as part of the API, rather than as part of a more user-facing quick-start guide.

I think this is a good question though, since we will probably add more documentation based on feedback from the upcoming review. Going to go ahead and merge, and if the organization of this page seems confusing we can discuss moving it somewhere else.

PR #198, opened 2023-12-30
Update copyright notice of docs index page

caleb531:
@eliotwrobson Just something I noticed and wanted to correct. I didn't change anything else, since it makes sense to me to remove the irrelevant README badges for the documentation site. Same for the links to the other documentation pages, as the links would probably break / navigate to the incorrect files anyway (we want to keep the user in the documentation site, after all).

One other thing: does `automata-lib[visual]` really require quotes to install?

```sh
pip install 'automata-lib[visual]'
```

...versus just:

```sh
pip install automata-lib[visual]
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 # Automata
 
-*Copyright 2016-2023 Caleb Evans*  
+*Copyright 2016-2024 Caleb Evans*  
 *Released under the MIT license*
 
 [![tests](https://github.com/caleb531/automata/actions/workflows/tests.yml/badge.svg)](https://github.com/caleb531/automata/actions/workflows/tests.yml)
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -1,7 +1,7 @@
 # Automata
 
-_Copyright 2016-2022 Caleb Evans_
-_Released under the MIT license_
+*Copyright 2016-2023 Caleb Evans*
```
eliotwrobson, line 3:
```suggestion
*Copyright 2016-2024 Caleb Evans*  
```
One day early lol

eliotwrobson (approved):
One small change.

Regarding the quotes in the install instructions, I believe those are actually required (I get an error when I don't use them).

docs/index.md
caleb531: @eliotwrobson Done in c58b594c1cfc92f6721df1224c3a5b8291e0e89d—but made sure to update all the other relevant copyright years as well.

eliotwrobson (approved):

PR #199, opened 2023-12-31
Redesign logo for smaller screen sizes

caleb531:
The favicon has also been regenerated based on this new logo. Let me know what you think! I wouldn't say it's a _huge_ difference, but then again, it's challenging to fit two states and a transition into one 24x24 icon lol.

Screenshot 2023-12-30 at 8 06 01 PM


eliotwrobson (changes requested):
I like the new logo overall! However, I think the color of the arrow is too similar to the background color of the docs site. Maybe something like orange that will contrast with the blue?

We should also put the new logo on the README.md

caleb531:
@eliotwrobson Yeah, you're right. I went with about the same blue as the original logo, but I agree it just doesn't work. I feel like orange might look a bit tacky—I was thinking for either white or some shade of yellow.

eliotwrobson:
@caleb531 I think yellow is a good choice!

caleb531:
@eliotwrobson Okay, the transition color is now yellow, and I think it's a big improvement indeed.

Screenshot 2023-12-30 at 8 14 29 PM

eliotwrobson (approved):
LGTM! I think adding the logo in more places can be a separate PR if you want.

caleb531:
@eliotwrobson Yeah, let's save that for a separate PR, since we'll have to think about layout concerns and all. Merging this now, though!

caleb531:
@eliotwrobson BTW, I've thought more about the merge-to-`main` thing—I was reminded that you committed some citation metedata to `main` directly without subsequently publishing a release, and I was fine with that because they were just housekeeping changes. So by extension, merges to `main` without publishing a release should be quite fine as well if they're just housekeeping changes like these.

With that, I've merged in the favicon changes to `main` so they can reflect on the docs site!
https://caleb531.github.io/automata/

Screenshot 2023-12-30 at 8 55 38 PM

PR #201, opened 2024-01-19
fix: Small changes from comment

eliotwrobson:
Some small changes suggested by this comment from the review: https://github.com/pyOpenSci/software-submission/issues/152#issuecomment-1900695651

Ran the build command locally and seems to work? Not 1000% sure if I tested this correctly, but I think fairly reasonable changes.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,7 +32,7 @@ jobs:
           pip install -r requirements.txt
 
       - name: Run mypy
-        run: mypy automata/**/*.py tests/*.py setup.py
+        run: mypy automata/**/*.py tests/*.py
 
       - name: Run flake8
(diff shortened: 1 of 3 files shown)


caleb531 (approved):
@eliotwrobson I read Leah's comment from the review thread, and this makes sense to me! It's also nice to know some of these practices are no longer necessary.

Approved!

caleb531:
@eliotwrobson BTW, feel free to merge these housekeeping changes into `main`. I also assume we don't need to publish another release given the nature of these changes.

caleb531:
@eliotwrobson But maybe first double-check that there aren't any other code changes that might be inadvertently merged in.

eliotwrobson:
@caleb531 merged to main! The only new commit was the one generated from this PR.

PR #204, opened 2024-01-31
Add Python 3.12 to CI matrix

caleb531:
@eliotwrobson This PR updates the CI jobs to run against GitHub Actions. It also adds runs the build and lint commands against multiple Python versions to ensure we aren't relying on any type features added since Python 3.8.

I _think_ I implemented everything correctly. It took a few tries to get CI working, but the issues were mostly fixed by upgrading certain dependencies in the toolchain.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,9 +4,9 @@ cached_method==0.1.0
 click==8.1.3
 coloraide==1.8.2
 coverage==6.4.4
-flake8==6.0.0
+flake8==7.0.0
 flake8-black==0.3.6
-flake8-isort==6.0.0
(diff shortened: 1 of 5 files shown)


requirements.txt
eliotwrobson, line 1: Do we want to try and update some of the other dev dependencies in this file (like black)?

eliotwrobson (approved):
One small question but otherwise seems good!

requirements.txt
caleb531: @eliotwrobson Probably a good idea, but I think I'd prefer to save that for a separate PR, if that makes sense to you. The intended scope for this PR was merely CI-related.

PR #205, opened 2024-02-01
Predecessor speed fix

### LINKED ISSUE #202: DFA: `maximum_word_length` complexity scales with number of final states
Since switching to networkx there seems to have been a degradation in speed for DFAs with a large number of final states.
For example I have one with 3348 final states and 11312 states in total.
The construction of coaccessible_nodes will run 3348 Breadth First Searches instead of running one with 3348 sources.
This results in the time complexity being O(NM) in the worst case.
This means that the check takes a full minute to run in my case which is not reasonable at this size.

Looking at networkx there isn't any direct way of changing our code to simply passing in multiple sources.
I'm honestly shocked that this isn't part of the interface.

https://github.com/caleb531/automata/blob/13e5c1b5bf1d92522eb83e914ad946552b77aaf7/automata/fa/dfa.py#L1656-L1658

eliotwrobson:
Resolves #202. @Tagl you can try this version of the code and see if it gives you the speedup you want.

### DIFF
--- a/automata/base/utils.py
+++ b/automata/base/utils.py
@@ -6,7 +6,7 @@ import os
 import pathlib
 import random
 import uuid
-from collections import defaultdict
+from collections import defaultdict, deque
 from itertools import count, tee, zip_longest
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


Tagl (approved):
This looks (practically) equivalent to the changes I made, but with fixes in additional places in `dfa.py`.

PR #208, opened 2024-02-11
Do not exclude any lines/branches from coverage report

caleb531:
@eliotwrobson With recent projects, I have had a change of perspective with regard to coverage. Rather than trying to achieve an often-unattainable 100% coverage score (even by excluding lines/branches that logically cannot be covered), I have reasoned it is better to give an honest and accurate score.

Because, of course, anyone can achieve a 100% code coverage score by ignoring lines/branches, and I would prefer to have the coverage scores in my projects be accurate reflections of the analysis, rather than cutting corners in the interest of "perfection".

Therefore, this PR removes all `# pragma: no cover` and `# pragma: no branch` directives, and updates the Contributing guide to remove any strict (or even implied requirement) for exact 100% coverage. Of course, we want as much to be covered as possible, but just as much as what can be achieved without ignoring lines/branches.

Anyway, that's my spiel. But I wanted to get your thoughts before adopting this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,8 +38,8 @@ increased from 79 characters to 120 characters.
 2. Whether you are introducing a bug fix or a new feature, you *must* add tests
 to verify that your code additions function correctly and break nothing else.
 
-3. Please run `coverage run -m nose2 && coverage report` to ensure that code
-coverage remains close to 100%.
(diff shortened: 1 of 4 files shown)


eliotwrobson (approved):
Seems reasonable! I think this is good to adopt before more changes from upcoming reviews.

Issue #213, opened 2024-03-23
demonstration of processing of large input


phildong:
The README mentioned optimized algorithm for processing of large inputs. It would be awesome to include some example/benchmarking demonstrating the capability in this regard.

caleb531:
That's a great idea! @eliotwrobson, is this something you would be able to work on?

eliotwrobson:
@caleb531 I should be able to take a look. The performance claims aren't with respect to any huge benchmarks or anything, so I can add something that randomly generates large examples for minimization to the docs.

phildong:
I think any quantitative benchmark would be helpful here.
Just want to give some specific details regarding what's "capable of processing large input", so people can at least have an idea and potentially compare to other packages.

eliotwrobson:
@phildong I have added the performance examples in #217. Please let me know what you think, or if there are other possible examples that you think might be helpful.

eliotwrobson:
Performance based examples have been added to the documentation in the develop branch! Closing.

PR #217, opened 2024-05-15
docs: Added the first examples of perf

eliotwrobson:
Work in progress performance documentation examples for #213. Will write second example shortly.

### DIFF
--- a/docs/examples/fa-examples.md
+++ b/docs/examples/fa-examples.md
@@ -1,4 +1,4 @@
-# FA Examples
+# Finite Automata Examples
 
 On this page, we give some short examples with discussion for the finite
 automata classes and methods.
--- a/docs/examples/perf-examples.md
+++ b/docs/examples/perf-examples.md
@@ -1,0 +1,78 @@
+# Performance Examples
+
+On this page, we give examples of reading large inputs with the library.
+We include timing code in these examples to allow for comparisons with
+other implementations.
+
+## Edit distance automaton for large dictionaries
+
+This is an extended version of the [edit distance example](fa-examples#edit-distance-automaton) on a large input. Here, we will
+return all words in the given english dictionary within the specified
+edit distance to the target word.
+
+```python
+# Imports from automata library
+import string
+import time
+
+import pooch
+
+from automata.fa.dfa import DFA
+from automata.fa.nfa import NFA
+
+# First, get a set of all the words we'd like to use
+word_file = pooch.retrieve(
+    url="https://raw.githubusercontent.com/solardiz/wordlists/master/gutenberg-all-lowercase-words.txt",
+    known_hash="62be81d8a5cb2dae11b96bdf85568436b137b9c0961340569ca1fca595774788",
+)
+
+with open(word_file, "r") as wf:
+    word_set = set(wf.read().splitlines())
+
+
+print("Word set size:", len(word_set))
+
+# Create the DFA recognizing all the words we'd like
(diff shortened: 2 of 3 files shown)


caleb531:
@eliotwrobson Is this ready for my review yet? I would ask that any work-in-progress PRs be kept in Draft state until they are ready for my final approval.

eliotwrobson:
@caleb531 not quite yet, this PR is in the unique situation where we're really waiting on approval by the reviewers that opened the original issue. They may ask for more changes, so makes sense to leave as a draft for now 👍

docs/examples/fa-examples.md
```diff
@@ -1,12 +1,27 @@
...
+# Finite Automata Examples
 
 On this page, we give some short examples with discussion for the finite
 automata classes and methods.
 
 ## Reading basic input
 
-The following code snippet creates a `DFA` and prints whether it accepts or
-rejects user input.
+In this example, we define a function that takes in an automata
```
caleb531, line 8:
```suggestion
In this example, we define a function that takes in an automaton
```

docs/examples/fa-examples.md
```diff
@@ -24,76 +39,113 @@ my_dfa = DFA(
...
-            print('Rejected')
-except KeyboardInterrupt:
-    print('')
+read_user_input(my_dfa)
+```
+
+We can also do the same with an NFA we define. Note that the
+transition dictionary for the NFA has a different structure than
+that of the DFA, and that we are working over a different input
+alphabet than the previous example:
```
caleb531, line 48: @eliotwrobson Where is the machine of the previous example defined? I think the user will assume the previous example is the REPL example above, which does not show the definition of `my_automaton` (and it doesn't necessarily need to, but I think this paragraph should be reworded to be clearer / more accurate.

docs/examples/fa-examples.md
```diff
@@ -24,76 +39,113 @@ my_dfa = DFA(
...
+    initial_state="q0",
+    final_states={"q1"},
+)
+
+read_user_input(my_nfa)
 ```
 
 ## Subset for NFAs
 
 The `NFA` does not have a built-in method for checking whether it is a subset
```
caleb531, line 72: @eliotwrobson Can you please remind me the reason we don't have an `issubset` method for NFAs? I feel like it was an intentional decision we made earlier this year, but I can't remember why.

docs/examples/perf-examples.md
```diff
@@ -0,0 +1,120 @@
+# Performance Examples
+
+On this page, we give examples of reading large inputs with the library.
+We include timing code in these examples to allow for comparisons with
+other implementations.
+
+## Edit distance automaton for large dictionaries
+
+This is an extended version of the [edit distance example](../fa-examples#edit-distance-automaton) on a large input. Here, we will
+return all words in the given english dictionary within the specified
```
caleb531, line 10: Totally a nitpick, but just something I noticed:

```suggestion
return all words in the given English dictionary within the specified
```

docs/examples/fa-examples.md
eliotwrobson: It would require creating a new NFA in-memory. In hindsight, maybe a little overkill to worry about, I think the overhead on performance wouldn't be too bad.

caleb531: @eliotwrobson Fair enough! We can revisit that discussion later.

caleb531 (approved):

PR #218, opened 2024-05-16
fix: Fix bug in minify edge case

eliotwrobson:
Fixes small bug in the minify function for partial DFAs and the as_partial check for a DFA creation function.

I think this raises the question, @caleb531 should the allow_partial be a computed property in the constructor if it is manually set by the user? i.e., if client code sets this property to true, should it be set to `False` if the transition dictionary is not partial.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -625,7 +625,9 @@ class DFA(fa.FA):
                                 x for x in count(-1, -1) if x not in reachable_states
                             )
                             for trap_symbol in input_symbols:
-                                transition_back_map[trap_symbol][trap_state] = [trap_state]
+                                transition_back_map[trap_symbol][trap_state] = [
+                                    trap_state
+                                ]
 
                             reachable_states.add(trap_state)
 
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -682,6 +682,11 @@ class DFA(fa.FA):
             if trap_state not in eq
         }
 
+        # If only one equivalence class with the trap state,
+        # return empty language.
+        if not back_map:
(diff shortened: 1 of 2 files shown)


caleb531:
@eliotwrobson If the user explicitly sets `allow_partial`, then the machine should honor that choice. I'd prefer not to traverse the entire transition map as a computed property—that feels expensive. I can see a use case where the user passes in a variable transition map which may or may not be partial (perhaps it depends on user input).

caleb531 (approved):

eliotwrobson:
@caleb531 that line of reasoning makes sense. The reason I bring this up is because the `to_complete` method relies on the `allow_partial` only being set to true if the DFA is indeed partial (otherwise it adds an extra state, which is what was causing the bug here in the first place):

https://github.com/caleb531/automata/blob/develop/automata/fa/dfa.py#L341

I think the way to go then is to do the check for being partial at this function call, instead of relying on this boolean property. What do you think? This discussion is a bit orthogonal to the point of this PR, so I'll go ahead and merge. Thanks for the fast approval!

caleb531:
@eliotwrobson Yes, I think that is the approach to take. For me, the `allow_partial` parameter ultimately represents user choice, not the guaranteed state of the machine (hence the name—`allow_partial`, not `is_partial`). Therefore, it should not be relied upon as an assumption of the actual state of the machine (using the term "state" loosely here).

I suppose in theory, we could drop `allow_partial` entirely and just compute at instantiation time whether the machine is partial or not. But again, that sounds rather expensive to me, and also makes current non-partial validation checks pointless (since effectively this would allow every machine to be partial and get away with it).

PR #224, opened 2024-05-19
Successors length limits

### LINKED ISSUE #207: Add `min_length` and `max_length` parameters to `successors`/`predecessors`
I think the `successors`/`predecessors` functions are quite useful for enumerating words that are accepted by the DFA.
Sometimes I want to limit the length of the word in an infinite language.
So far I have been limiting lengths by intersecting with `DFA.of_length`.

It might be a good idea to add parameters for this, as it may be far more efficient to take apply that approach than to intersect with another DFA.
One additional benefit of this is it will guarantee that execution of the function ends, which is not true right now for infinite languages.
We might even want to require `max_length` or default to some reasonable upper bound based on the size of the alphabet.
I'm planning on changing the implementation myself to support this in the near future, assuming it is beneficial with regards to performance.

@caleb531 @eliotwrobson do you have any thoughts on this?

Tagl:
Closes #207 

Might be wise to add a few more tests.

### DIFF
--- a/automata/fa/dfa.py
+++ b/automata/fa/dfa.py
@@ -1276,6 +1276,8 @@ class DFA(fa.FA):
         *,
         strict: bool = True,
         key: Optional[Callable[[Any], Any]] = None,
+        min_length: int = 0,
+        max_length: Optional[int] = None,
     ) -> Optional[str]:
         """
         Returns the first string accepted by the DFA that comes before
@@ -1291,6 +1293,10 @@ class DFA(fa.FA):
         key : Optional[Callable], default: None
             Function for defining custom lexicographical ordering. Defaults to using
             the standard string ordering.
+        min_length : int, default: 0
+            Limits generation to words with at least the given length.
+        max_length : Optional[int], default: None
+            Limits generation to words with at most the given length.
 
         Returns
         ------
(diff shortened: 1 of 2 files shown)


eliotwrobson:
Looks good from a glance! When I make a pass through the active PR list a little later I can add some tests before merging.

caleb531:
@eliotwrobson @Tagl Looks like a great addition to me! Thank you for this!

I am converting the PR to Draft state until more tests are added.

eliotwrobson:
@Tagl added a couple of test cases and encountered a bug I commented on:

https://github.com/caleb531/automata/pull/224/commits/f95b0d17852cce7e5e8239fae83481d9d7f0da92#diff-404c46da3c65463b61109992f94cfd102f9f20f5467f1916bf9bf50ff89c7176R1894-R1895

I think this can be avoided just by popping from the char stack if the size goes over the length limit.

eliotwrobson:
@Tagl for this, should the stop condition be changed to when the algorithm hits the string of all the lexicographically last character of length `max_length`? I think that might fix the issue I found in my testing.

Tagl:
Should be fixed now, I added one more edge case to test for as well.

eliotwrobson (approved):
Looks good now that we have that edge case sorted out, thanks! Going to go ahead and approve. @Tagl do you think this is good to merge now?

Tagl:
Yes, I believe so.

Issue #232, opened 2024-07-18
Unicode characters with regexp ?


kunchtler:
Hello all,
Is there a reason why general unicode characters are not usable with regexp ?
I'm using this library to compute the regular expressions of automata with large alphabet (> 100 letters sometimes) so it would be nice to have this feature.
(I seem to have it working on my side by changing a single line of code with all tests still passing)
Thanks !

eliotwrobson:
This had to do with the regex we used to do parsing. If you have a change that enables using any Unicode character, we'd be happy to review and accept a pull request!

eliotwrobson:
PR was merged to develop, should go out with the next deploy.

PR #233, opened 2024-07-18
Unicode characters support in regex

kunchtler:
Linked to #232.

Changed the order in which the tokens are registered in the regex lexer to process the rule about recognizing letters last, and changed that rule to account for all non-blank characters (as specified per python's `re` library with `\S`).

Added a test to check for the support of non-ascii characters.

This is my very first pull request ever so feel free to guide me.

### DIFF
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -209,7 +209,7 @@ class NFA(fa.FA):
             The regex to construct an equivalent NFA for.
         input_symbols : Optional[AbstractSet[str]], default: None
             The set of input symbols to create the NFA over. If not
-            set, defaults to all letters found in the regex.
+            set, defaults to all non-reserved characters found in the regex.
 
         Returns
         ------
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -2,7 +2,6 @@
 """Classes and methods for working with nondeterministic finite automata."""
 from __future__ import annotations
 
-import string
 from collections import deque
 from itertools import chain, count, product, repeat
 from typing import (
--- a/automata/fa/nfa.py
+++ b/automata/fa/nfa.py
@@ -37,7 +37,6 @@ NFAStateT = fa.FAStateT
 NFAPathT = Mapping[str, AbstractSet[NFAStateT]]
 NFATransitionsT = Mapping[NFAStateT, NFAPathT]
 InputPathListT = List[Tuple[NFAStateT, NFAStateT, str]]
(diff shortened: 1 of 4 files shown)


eliotwrobson:
@kunchtler thanks for this! As part of the test, could you check the dictionary of input symbols matches the unicode characters used in the expression? Other than that, I think the test + new code looks good!

tests/test_regex.py
```diff
@@ -43,6 +43,10 @@ def test_helper_validate_invalid(self) -> None:
         """Should pass validation for valid regular expression"""
         self.assertFalse(re._validate("a(|)"))
 
+    def test_validate_unicode_characters(self) -> None:
+        """Should pass validation for regular expressions with unicode characters."""
+        re.validate("(µ|🤖ù)*")
+
```
eliotwrobson, line 49: Should add a test that an nfa converted from this regex has the expected set of input symbols.

eliotwrobson (changes requested):
@kunchtler thanks for this! One request to make this test a little more robust, but overall I think the change looks good.

tests/test_regex.py
kunchtler: Hello,

I added a test for this (not sure if it it should be in `test_regex.py` or `test_nfa.py`, so I put it in the former).

I also changed a bit of `NFA.from_regex(regex, input_symbols)` (which could be reverted, or be its own issue).
Currently, the default behaviour for when input_symbols is None is to default to the set of ascii lowercase letters and numbers.
Wouldn't it make more sense if it rather defaulted to all letters present in the regex ?

eliotwrobson: I think `test_regex` is a fine location to put this test.

For the second item, the prior default behavior made sense when the allowable alphabet was small, but now allowing all unicode characters, I think your choice of a default makes sense 👍🏽

automata/fa/nfa.py
```diff
@@ -211,7 +209,7 @@ def from_regex(
             The regex to construct an equivalent NFA for.
         input_symbols : Optional[AbstractSet[str]], default: None
             The set of input symbols to create the NFA over. If not
-            set, defaults to all ascii letters and digits.
+            set, defaults to all letters found in the regex.
```
eliotwrobson, line 212:
```suggestion
            set, defaults to all non-reserved characters found in the regex.
```
Now that we can have arbitrary unicode, we should talk about this in terms of characters and not letters.

eliotwrobson (approved):
LGTM with the new test!

PR #234, opened 2024-07-27
Added initial DFA notebook.

eliotwrobson:
Addition of self-contained example notebooks explaining finite automata theory using the library. More complete version of #212. Only the DFA notebook so far. Math isn't rendering because of this issue: https://github.com/danielfrg/mkdocs-jupyter/issues/143

We may have to wait till that's fixed before merging. 

cc @skyien-z


eliotwrobson:
@caleb531 how do you think we should go about adding this? It doesn't look like we can really add this notebook to the docs site until the math display is working, but I think the github jupyter notebook browser is sufficient. Should we just add a link to the docs site?

caleb531:
@eliotwrobson Yeah, I think just adding a link to the Docs site for now is fine.

eliotwrobson:
@caleb531 sounds good! Added the link and will go ahead and merge, since I edited the notebook earlier.

PR #238, opened 2024-11-08
Updated settings.json + ms-python.isort as recommended extension

CamiloMartinezM:
Since we're using Flake8 and black, we should put the 88-line ruler, which is black's default maximum line length, so I added `"editor.rulers": [88]` to `.vscode/settings.json`.

image

 On the other hand, `"isort.args"` is only available if we install [ms-python.isort](https://marketplace.visualstudio.com/items?itemName=ms-python.isort), which is why I added it to `.vscode/extensions.json`.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,6 +2,7 @@
   "recommendations": [
     "EditorConfig.EditorConfig",
     "ms-python.black-formatter",
+    "ms-python.isort",
     "matangover.mypy",
     "ms-python.flake8"
   ]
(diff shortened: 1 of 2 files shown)


caleb531:
Seema reasonable to me. @eliotwrobson, any objections?

Also, @CamiloMartinezM, for future PRs, please always set the target branch as `develop`, not `main`.

eliotwrobson:
@caleb531 go for it! Sorry I've been a little absent, very busy as of late.

caleb531 (approved):

PR #243, opened 2025-01-19
Ruff + 3.13 version bump

### LINKED ISSUE #242: Switch formatter to ruff
See title, can replace flake8, black, and isort, and it's way faster: https://docs.astral.sh/ruff/

eliotwrobson:
Resolves #242 

@caleb531 if you decide to merge this, it's probably time for a v8.4.1 release (since we added a couple small PRs on develop). I'm happy to write the release notes.


pyproject.toml
eliotwrobson, line 1: Some additional changes here because I used a toml autoformatter.

pyproject.toml
```diff
@@ -56,22 +62,31 @@ disallow_untyped_defs = true
...
-# Black compatibility
-max-line-length = 88
-extend-ignore = ["E203", "W503"]
+[tool.ruff]
+include = ["**/*.py"]
 
-# Per <https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#profile>
-[tool.isort]
-profile = "black"
+[tool.ruff.lint]
```
eliotwrobson, line 76: This is a pretty minimal ruleset, but there are many others that can be enabled to check things like naming conventions, warnings against outdated programming patterns, etc: https://docs.astral.sh/ruff/rules/

docs/migration.md
```diff
@@ -4,7 +4,7 @@
 
 ### Dependency Changes
 
-Python 3.7 support has been dropped. Please upgrade to Python 3.8 or later to
+Python 3.7 support has been dropped. Please upgrade to Python 3.9 or later to
```
caleb531, line 7: @eliotwrobson Why did this change? This seemed logically correct before (i.e. if you drop 3.7 support, then 3.8 is the minimum supported version, not 3.9).

eliotwrobson: We just waited too long to update this: https://devguide.python.org/versions/

caleb531: @eliotwrobson Yes, but even if 3.8 hits EOL, dropping support for it still necessitates a semver-major version bump because you it technically constitutes a breaking change for users still on Python 3.8 for whatever reason.

Not that I would be opposed to dropping 3.8 support now that it _is_ EOL, but I would prefer that the release version accurately reflects the drop in support. With that, I might suggest we hold off on this PR until we release what has already been merged (i.e. to make for the nice 8.4.1 release you mentioned), then release v9 with drop for 3.8 support.

Thoughts?

eliotwrobson: I think that makes sense, though doing multiple consecutive releases seems a little overkill for something like this. I'd rather just release a minimal 9.0 with everything we have so far (I think develop only has one additional feature that is just expanded unicode support and a bugfix).

caleb531: @eliotwrobson I'm inclined to agree, actually. Let's just stick to a single v9.0.0 release.

As far as updating the migration guide, I think I'd still recommend leaving at 3.8, though. Because otherwise, are we setting a precedent where we feel obligated to keep track of core Python's EOL timeline and update all past Python-version-dropping major releases in the migration guide?

eliotwrobson: I agree, I'll go ahead and add a new v9 section 👍🏽

pyproject.toml
caleb531: @eliotwrobson I think what you have is fine, as long as it generally covers what Flake8+black+isort already cover out of the box.

pyproject.toml
caleb531: @eliotwrobson Will ruff not autoformat the TOML for us? What autoformatter did you use? Was it automated or did you run it manually?

pyproject.toml
eliotwrobson: Yup, it does that + more than what we had before.

docs/migration.md
eliotwrobson: Added!

pyproject.toml
eliotwrobson: Ruff doesn't handle this (only the .py files). I used the autoformatter here: https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml

Maybe we should add to the recommended extensions?

caleb531: @eliotwrobson That's fair. I took a look at the extension's listing page, and it considers itself a "preview extension" that may not represent something entirely stable. So I'm not sure it it's worth designating as an official recommendation, especially given the limited value-add.

eliotwrobson: @caleb531 makes sense and I agree. Should I go ahead and merge this?

caleb531: @eliotwrobson I'm just giving everything one more look over. But we're pretty close. I might just have a few additional questions.

.vscode/settings.json
```diff
@@ -2,14 +2,11 @@
   "python.analysis.extraPaths": [".virtualenv"],
   "[python]": {
     "editor.formatOnSave": true,
-    "editor.defaultFormatter": "ms-python.black-formatter",
+    "editor.defaultFormatter": "charliermarsh.ruff",
     "editor.rulers": [88],
     "editor.codeActionsOnSave": {
+      "source.fixAll": "explicit",
```
caleb531, line 8: @eliotwrobson I'm guessing this enables Ruff to fix sort order on file save? Per https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff:

> You can configure Ruff to fix lint violations on-save by enabling the source.fixAll action in settings.json:
> ```json
> {
>   "[python]": {
>     "editor.codeActionsOnSave": {
>       "source.fixAll": "explicit"
>     }
>   }
> }
> ```

I'm surprised this is necessary since the [isort VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-python.isort) was able to update files automatically without the need for additional configuration.

.vscode/extensions.json
```diff
@@ -1,9 +1,7 @@
 {
   "recommendations": [
+    "charliermarsh.ruff",
```
caleb531, line 3: @eliotwrobson In your experience, how does the Ruff extension work for you compared to Flake8? Similar experience?

eliotwrobson: Much better, it runs super fast and there's no reason to fuss with any plugins. I habitually switch to it on all my projects because of this.

.vscode/settings.json
eliotwrobson: Ahh this is because the ruff formatter and ruff linter are slightly different. Adding this enable fixing lint violations on save, but that's different than the formatter itself (I think).

.vscode/extensions.json
caleb531: @eliotwrobson Honestly, I was considering switching my projects to use Ruff as well. it sounds pretty great, and I like the idea of consolidating my dev dependencies.

Just to do my due diligence, I'm going to install the VS Code extension under this PR to give it a try before we merge.

eliotwrobson: It's honestly a great experince to use so I think it's worth the time 😄

.vscode/settings.json
caleb531: @eliotwrobson Fair enough. I would think that formatting violations technically constitute linting violations (sort of how they do in flake8 and flake8-black, e.g. the "Black would require changes" linter message).

Let's keep it, though! We want to have compliant and quality code in the codebase, and enforcing it through automation (i.e. `source.fixAll`) helps maintain that standard.

.vscode/extensions.json
caleb531: @eliotwrobson Wow, that is blazing fast. So fast that I'm getting syntax errors from Ruff before I can finish typing out a new line of code 🤣. But that's fine! I don't think that will get in the way of development.

With that, I think this PR is good to go, and I will go ahead and merge.

caleb531 (approved):

PR #250, opened 2025-03-06
Add support for character classes `[...]`

### LINKED ISSUE #249: Regex: Support for character classes `[...]`
What would be required to add support for character classes?

StefanosChaliasos:
Fixes #249 

It also makes the quantifiers more expressive:

I.e., now it supports: {,4}, {4}, {1,3}, {1,} instead of just {1,3} and {1,}

### DIFF
--- a/tests/test_regex.py
+++ b/tests/test_regex.py
@@ -428,7 +428,10 @@ class TestRegex(unittest.TestCase):
             cyrillic_nfa.accepts_input("\u0401a")
         )  # Mixed with non-matching
 
-        combined_regex = "Latin-Extension[\u00a1-\u01bf]+Greek[\u0370-\u03ff]+Cyrillic[\u0400-\u04ff]+"
+        combined_regex = (
+            "Latin-Extension[\u00a1-\u01bf]+Greek[\u0370-\u03ff]+"
+            "Cyrillic[\u0400-\u04ff]+"
+        )
         combined_nfa = NFA.from_regex(combined_regex, input_symbols=input_symbols)
 
         self.assertTrue(


StefanosChaliasos:
Will go over everything tomorrow. Thanks a lot for the feedback.

StefanosChaliasos:
I did some more changes, can you review the new ones. Basically I added support for shorthand (e.g., '\d') and I tokenised whitespace. I need to add more tests and polish the code. I'll change the PR as a draft until done.

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
+        if "\\S" in regex:
+            additional_symbols.update(NON_WHITESPACE_CHARS)
+        if "\\d" in regex:
+            additional_symbols.update(DIGIT_CHARS)
+        if "\\D" in regex:
+            additional_symbols.update(NON_DIGIT_CHARS)
+        if "\\w" in regex:
+            additional_symbols.update(WORD_CHARS)
+        if "\\W" in regex:
+            additional_symbols.update(NON_WORD_CHARS)
```
caleb531, line 245: @StefanosChaliasos Can you please refactor this to use a `dict`-based lookup table? That would make this much less repetitive.

cc @eliotwrobson

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
         """
+        # Import the shorthand character classes
+        from automata.regex.parser import (
+            DIGIT_CHARS,
+            NON_DIGIT_CHARS,
+            NON_WHITESPACE_CHARS,
+            NON_WORD_CHARS,
+            WHITESPACE_CHARS,
+            WORD_CHARS,
+        )
```
caleb531, line 228: @StefanosChaliasos Can you please keep all imports at the top of the file? There's no particular need for the tighter scoping here, IMO.

cc @eliotwrobson

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
+                        continue
+                    elif class_content[pos + 1] == "d":
+                        additional_symbols.update(DIGIT_CHARS)
+                        pos += 2
+                        continue
+                    elif class_content[pos + 1] == "w":
+                        additional_symbols.update(WORD_CHARS)
+                        pos += 2
+                        continue
+                    elif class_content[pos + 1] in "S":
```
caleb531, line 285: @StefanosChaliasos What is the intention of using `in` here as opposed to `==`? If the right-hand side is just a single character, the only difference that seems to make is permitting `class_content[pos + 1]` to be empty string (in addition to the character itself). In other words:

```
"S" in "S" True
"" in "S"  # True
```

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
+                        additional_symbols.update(NON_DIGIT_CHARS)
+                        pos += 2
+                        continue
+                    elif class_content[pos + 1] in "W":
+                        additional_symbols.update(NON_WORD_CHARS)
+                        pos += 2
+                        continue
+
+                    # Handle escape sequence in character class
+                    from automata.regex.parser import _handle_escape_sequences
```
caleb531, line 299: @StefanosChaliasos Can you also please move this import to the top of the file?

automata/regex/parser.py
```diff
@@ -24,10 +25,21 @@
     validate_tokens,
 )
 
+# Add these at the top of the file to define our shorthand character sets
+ASCII_PRINTABLE_CHARS = frozenset(string.printable)
```
caleb531, line 29: @eliotwrobson The implication here is that only ASCII characters are deemed as printable characters, but how would that work given that #233 just added support for Unicode characters?

caleb531:
Hey, @StefanosChaliasos! I left some additional comments on the PR—apologies if they seem nitpicky, but just wanting to maintain solid code quality and consistency for this project.

StefanosChaliasos:
Thanks for the review, I will address the comments once I find some time

automata/regex/parser.py
eliotwrobson: I think this only gets used when adding characters from the use of some special character classes? It might be good to have a test case that uses a non-printable character.

automata/fa/nfa.py
eliotwrobson: I agree this could make things much cleaner, especially since this can be done in a loop 👍🏽

automata/fa/nfa.py
eliotwrobson: Yes, I think ruff will complain about the imports.

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
+
+        class_symbols = set()
+        range_pattern = re.compile(r"\[([^\]]*)\]")
+        for match in range_pattern.finditer(regex):
+            class_content = match.group(1)
+            pos = 0
+            while pos < len(class_content):
+                if class_content[pos] == "\\" and pos + 1 < len(class_content):
+                    # Check for shorthand classes in character classes
+                    if class_content[pos + 1] == "s":
```
eliotwrobson, line 273: I think you might be able to use a lookup table from a dictionary here? Just use the character as a key and a tuple of additional symbols and position increment as the value.

automata/fa/nfa.py
```diff
@@ -216,21 +217,165 @@ def from_regex(
...
+                        ):
+                            end_char = _handle_escape_sequences(class_content[pos + 4])
+                            pos += 5
+                        else:
+                            end_char = class_content[pos + 3]
+                            pos += 4
+
+                        # Add all characters in the range to input symbols
+                        for i in range(ord(start_char), ord(end_char) + 1):
+                            class_symbols.add(chr(i))
```
eliotwrobson, line 325: I think this can be done with a python `update` call instead of a loop.

automata/regex/parser.py
```diff
@@ -548,18 +659,116 @@ def get_regex_lexer(
...
+        "[": "[",
+        "]": "]",
+        "{": "{",
+        "}": "}",
+        "^": "^",
+        "$": "$",
+        "&": "&",
+    }
+
+    if char in escape_map:
```
eliotwrobson, line 767:
```suggestion
    return escape_map.get(char, char)
```

eliotwrobson (commented):
A few more comments. @StefanosChaliasos also, we just updated the develop branch to use UV, so be sure to do a rebase before adding more changes. It would be awesome if we could close this out soon, since along with the switch to UV, we have a couple of feature request PRs that could be part of a new release.

StefanosChaliasos:
Thanks for the additional comments, will fix everything by the end of the week. Got busy with other stuff :)

eliotwrobson:
@StefanosChaliasos just want to check in briefly since this will now require a rebase on the develop branch because of some changes that have been made on our end. The comments I made earlier are relatively minor, so let me know if you'd like any help landing this. I think we want to cut a new release before too long so it would be awesome to have these changes included 👍🏽

caleb531 (approved):
@eliotwrobson Looks good to me! Approved—please merge at your earliest convenience.

PR #252, opened 2025-03-18
Added Function of Generating Animation of DFA and NFA Reading Strings

dsr20030703:
I have added a `animate_reding_input` function in DFA and NFA to automatically generate the animation of the DFA or NFA reading the input using [`manim` 0.19.0](https://pypi.org/project/manim/).

I have also added a Jupyter Notebook for the funciton in the `example_notebooks/` folder to show the generated animation.

As there're some problems in the type lint of `manim` and `pygraphviz`, which includes:
1. Argument `color` to `manim.FadeToColor` expected `str`, but it also accepts type `ManimColor`;
2. The docstring of `pygraphviz.Node` and `pygraphviz.Edge` says each node or edge has attributes that can be directly accessed through the `attr` dictionary, but they aren't recognized by mypy.

It comes with some errors in typechecking run with mypy. But there aren't any functional bugs (as far as I have checked). 

Another problem I have encountered is that I don't know how to add the tests. As `manim` is an animation engine, they have developed [a unique test framework based on pytest](https://docs.manim.community/en/stable/contributing/testing.html), which is different from this project. So I haven't added any test. (But at least the examples in the Jupyter Notebook all works, except the last of which)


.gitignore
```diff
@@ -24,3 +24,6 @@ nosetests.xml
 
 # MyPy
 .mypy_cache
+
+# manim generated files
+media/
```
eliotwrobson, line 29: Should have a newline at the end of the file.

automata/base/animation.py
```diff
@@ -0,0 +1,255 @@
...
+import manim
+import pygraphviz as pgv
+
+
+class Animate:
+    """The `Animate` class is the config class to set the behavior of animations."""
+
+    DEFAULT_COLOR = manim.WHITE
+    HIGHLIGHT_COLOR = manim.RED
+    manim.Circle.set_default(color=DEFAULT_COLOR)
```
eliotwrobson, line 13: Does this change a global property of `manim`? Is there a way to do this local to what is animated from this class?

automata/base/animation.py
```diff
@@ -0,0 +1,255 @@
...
+        - Circle for non-final state
+        - VGroup of a doublecircle for final state
+    label: Optional[Text]
+        The name of the state (N/A for the nullnode)
+    """
+
+    shape: manim.Dot | manim.Circle | manim.VGroup
+    label: Optional[manim.Text]
+
+    def __init__(self, node: pgv.Node):
```
eliotwrobson, line 69:
```suggestion
    def __init__(self, node: pgv.Node) -> None:
```

automata/base/animation.py
```diff
@@ -0,0 +1,255 @@
...
+                    circle,
+                    manim.Circle().surround(circle, buffer_factor=0.8),
+                )
+                if node.attr["shape"].startswith("double")
+                else circle
+            )
+            self.add(self.shape)
+            self.label = manim.Text(node.name, font_size=float(node.attr["fontsize"]))
+            self.add(self.label)
+        x, y = (float(pt) / 72 for pt in node.attr["pos"].split(","))
```
eliotwrobson, line 102: 72 is a magic number here, could this be set at the class level or at the top of the file?

automata/base/animation.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+        return result
+
+
+class _ManimInput(manim.VGroup):
+    """The `ManimInput` class represents the input string. The characters of the input
+    string are separated with each character is generated to a `Text` object, so that
+    you can get each character simply with `[]` operator."""
+
+    def __init__(self, text: str):
```
eliotwrobson, line 188:
```suggestion
    def __init__(self, text: str) -> None:
```

automata/fa/dfa.py
```diff
@@ -2536,3 +2561,92 @@ def _get_input_path(
         accepted = last_state in self.final_states
 
         return path, accepted
+
+
+if fa._visual_imports:
+
+    @dataclass(eq=False)
```
eliotwrobson, line 2568: Should we put this in the `animation` file, or make a separate folder for just animations? I'm not a fan of having an entire class definition behind an if statement like this, but not 100% sure what the best way to do this is. @caleb531 any thoughts on this?

example_notebooks/fa_animation.ipynb
eliotwrobson, line 1: Thanks for adding the example notebook!

eliotwrobson (commented):
Several comments. My main reservation is having certain class definitions behind if statements, but the code itself seems solid. Should consult with @caleb531 to see what the best way to organize these changes is. Overall looks good though, thanks again for this contribution!!

automata/base/animation.py
dsr20030703: Yes, it changes the default color of `manim.Circle` from `manim.RED` to `manim.WHITE` globally. A way to do this local to what is animated from this class is by specifying `color=Animate.DEFAULT_COLOR` in every `manim.MObject` created, which is what I did at first. After that, I thought it may be too troublesome. As most of `manim.MObject` except `manim.Circle` uses `manim.WHITE` as default color, I chose to change the default color of `manim.Circle` to `manim.WHITE` globally.

Is the way of specifying `color=Animate.DEFAULT_COLOR` in every `manim.MObject` created suggested? If so, I'll revise the code in that way.

automata/base/animation.py
dsr20030703: 72 is the ratio between inch and point (actually, big boint). Graphviz measures some attributes of Graph, Node and Edge in points, while some attibutes are measured in inches.

Is adding a constant like `_POINTS_IN_INCH = 72` at the top of the file, and replacing all "72" with "_POINTS_IN_INCH" OK?

automata/fa/dfa.py
dsr20030703: As `_DFAAnimation` and `_NFAAnimation` requires `DFA` and `NFA` respectively, and the `animate_reading_input` function in `DFA` and `NFA` needs to call the initialize function of `_DFAAnimation` and `_NFAAnimation` respectively, it seems that circular dependencies can't be avoided. :( If we put them into other files, like a new `automata/fa/animation.py` file, then we have to put the import statement inside the `animate_reading_input` function to prevent circular import, like this:
```py
class DFA:
    ...
    def animate_reading_input(self, input_str: str, preview: bool = False) -> None:
        if not fa._visual_imports:
            raise ImportError(...)
        from automata.fa.animation import _DFAAnimation

        _DFAAnimation(self, input_str).render(preview)
```
Is that acceptable?

automata/base/animation.py
eliotwrobson: Yes, I think that's a good strategy 👍🏽

automata/fa/dfa.py
eliotwrobson: Let me think about this some and get back to you. Circular dependencies like these can usually be resolved by changing the design somehow.

Also, I think the circular dependency may only be because of the DFA / NFA type annotation? In that case, you can put those imports behind an `if TYPE_CHECKING` guard, see [here](https://vickiboykis.com/2023/12/11/why-if-type_checking/) for discussion.

automata/base/animation.py
dsr20030703: Is a helper function like this OK? (`partial` is from `functools`)
```suggestion
    M = TypeVar("M", bound=manim.Mobject)

    @classmethod
    def default_init(cls, mobject_class: type[M]):
        return partial(mobject_class, color=cls.DEFAULT_COLOR)
```
It can handle all subclasses of `manim.Mobject` like `manim.Circle`, `manim.Text`, etc. In that case, we would use `Animate.default_init(/*a manim.Mobject class*/)` to replace the default init function of that class. For example, `manim.Text(node.name, font_size=float(node.attr["fontsize"]))` will be replaced by `Animate.default_init(manim.Text)(node.name, font_size=float(node.attr["fontsize"]))`, and `manim.Circle()` will be repalced by `Animate.default_init(manim.Circle)()` (seems that the latter `()` are easy to forget though).

automata/fa/dfa.py
dsr20030703: > Also, I think the circular dependency may only be because of the DFA / NFA type annotation? In that case, you can put those imports behind an `if TYPE_CHECKING` guard, see [here](https://vickiboykis.com/2023/12/11/why-if-type_checking/) for discussion.

It works! Now we don't need to import the module when calling.

automata/base/animation.py
eliotwrobson: I think that seems reasonable! My main motivation for this comment is to avoid side effects that can cause other animation code to behave unexpectedly, and I think this is a solid way of doing that.

.gitignore
dsr20030703: I've added a new line at the end of this file in the newest commit (6149742). And Git at local also has recorded the change. But I don't know why it's missed when uploading to the GitHub. 😕❓
I've added a new line
Git at local also has recorded the change

eliotwrobson: I think you have to configure VSCode to add newlines when saving any file, otherwise you might see weird behavior like this.

automata/fa/nfa.py
```diff
@@ -1275,3 +1278,22 @@ def _get_input_path(self, input_str: str) -> Tuple[InputPathListT, bool]:
...
+        ----------
+        input_str : str
+            The input string to read.
+        preview : bool, default: False
+            If true, opens scene in a file viewer.
+        """
+        if not fa._visual_imports:
+            raise ImportError(
+                "Missing visualization packages; "
+                "please install pygraphviz, coloraide, and manim."
```
eliotwrobson, line 1297: I think you still need to add `manim` to the visual imports section in `fa.py`. I think it would also be good to add a helper function that checks for the visual imports and raises an exception if they are not present (i.e. just put this if block into a helper function in `fa.py`).

eliotwrobson (changes requested):

automata/fa/nfa.py
dsr20030703: In spite of that `manim` isn't used in `fa.py`? Plus, I don't think that checking for the visual imports every time when calling them is a good idea, as the environment won't change when the program is running. Perhaps setting a variable recording the result of checking for the visual imports [using `importlib`](https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported) in the `automata.base.configs` would be better? In that case, we don't need to "try-import-except `ImportError`" to check visual imports every time, just coding
```py3
if config._visual_imports:
    import ...
```
is OK. Plus, we could let the `_visual_imports` to be a `Optional[ImportError]` object. When visual imports exists, its value is `None`; while when visual imports doesn't exist, its value is `ImportError("Missing visualization packages; please install ...")` (while in that case, it may be named as `miss_visual_imports`). So that we don't need to change all the notes of `ImportError`s when we want to change the visualization packages.

eliotwrobson: @dsr20030703 I think that idea makes sense. My suggestion was mainly to try to avoid repeating this message, so using the optional import error would accomplish this 👍🏽

eliotwrobson:
@dsr20030703 the overall structure of the code looks good now, but it looks like tests are failing. If you can get the tests passing, then I can do a full review 👍🏽

dsr20030703:
It seems that checks failed when installing dependencies because [some additional dependencies are required to install on Linux](https://docs.manim.community/en/stable/installation/uv.html#step-3-installing-manim). Should I add a new step after "Setup graphviz" in `.github/workflows/build.yml` like this?
```yml
- name: Build Additional Dependencies for manim
  run: |
    sudo apt update
    sudo apt install -y build-essential python3-dev libcairo2-dev libpango1.0-dev
```

eliotwrobson:
@dsr20030703 go for it 👍

requirements.txt
eliotwrobson, line 1: Missing a newline, and could you add the new dependency in alphabetical order with the rest?

caleb531 (commented):
@eliotwrobson @dsr20030703 Apologies for the very late response from me.

I'm gonna be completely honest, but my initial impression is that animation feels out-of-scope for this package. Looking over the PR, it feels like it's going to add a lot of weight that I think would be better suited for a fork. However, I am open to discussion on this point.

Also, out of curiosity, can the output directory be customized? It feels rather restrictive to have the generated files be restrictively written to a `media/` directory.

dsr20030703:
@caleb531 
> I'm gonna be completely honest, but my initial impression is that animation feels out-of-scope for this package. Looking over the PR, it feels like it's going to add a lot of weight that I think would be better suited for a fork. However, I am open to discussion on this point.

Will adding a new optional-dependencies like `animation` be acceptable? To be honest, the result of current `DFA.show_diagram` and `NFA.show_diagram` may suck when it's provided with a long `input_str` argument. For example, the code following
```py
dfa = DFA(
    states={"q0", "q1", "q2", "q3"},
    input_symbols={"0", "1"},
    transitions={
        "q0": {"0": "q2", "1": "q1"},
        "q1": {},
        "q2": {"0": "q3", "1": "q1"},
        "q3": {"0": "q2", "1": "q1"},
    },
    initial_state="q0",
    final_states={"q3"},
    allow_partial=True,
)
dfa.show_diagram("1111111111111")
```
will produce such a picture
dfa

> Also, out of curiosity, can the output directory be customized? It feels rather restrictive to have the generated files be restrictively written to a `media/` directory.

Yes, there're [a lot of configurations](https://docs.manim.community/en/stable/guides/configuration.html#a-list-of-all-cli-flags) in manim. Users can customize the output directory with `manim render --custom_folders`.

caleb531:
> Will adding a new optional-dependencies like `animation` be acceptable? The result of the current visualization function, `FA.show_diagram`, when provided with a long `input_str`, frankly speaking, may suck. For example, the following code
> ...
> Yes, there're [a lot of configurations](https://docs.manim.community/en/stable/guides/configuration.html#a-list-of-all-cli-flags) in manim. Users can customize the output directory with `manim render --custom_folders`.

Is it truly an optional dependency, though? Because I also saw your comment about needing to globally install Manim😕

> It seems that checks failed when installing dependencies because [some additional dependencies are required to install on Linux](https://docs.manim.community/en/stable/installation/uv.html#step-3-installing-manim). Should I add a new step after "Setup graphviz" in `.github/workflows/build.yml` like this?
> 
> ```yaml
> - name: Build Additional Dependencies for manim
>   run: |
>     sudo apt update
>     sudo apt install -y build-essential python3-dev libcairo2-dev libpango1.0-dev
> ```

I am mostly concerned about how people who don't need the animation functionality will potentially be required to install these sub-dependencies (or global dependencies, whatever you want to call them). But if we are able to work around this through the use of optional dependencies / addons (like we have the `[visual]` addon already), then I'd still consider myself open to this functionality.

cc @eliotwrobson

eliotwrobson:
> I am mostly concerned about how people who don't need the animation functionality will potentially be required to install these sub-dependencies (or global dependencies, whatever you want to call them). But if we are able to work around this through the use of optional dependencies / addons (like we have the `[visual]` addon already), then I'd still consider myself open to this functionality.

I could be wrong, but I think the way this PR is set up, the global dependencies are only needed if you're using the animation (i.e. someone could just install automata without the visual extra and everything would work fine), but we need to install everything in CI to run tests. I definitely agree though that there should be a way to make this additional functionality optional.

dsr20030703:
It seems that `build.yml`, `lint.yml` and `tests.yml` shall be changed simultaneously when adding new dependencies. As the first 4 steps of them are same, could those steps be extracted and shared by 3 jobs so that just 1 place need to be changed?

Also, it seems that [this project is going to adopt  uv](#254 )? It would help a lot in setting up since [Manim also recommends using uv](https://docs.manim.community/en/stable/installation/uv.html). Shall I wait until #254 gets reviewed and merged?

caleb531:
> It seems that `build.yml`, `lint.yml` and `tests.yml` shall be changed simultaneously when adding new dependencies. As the first 4 steps of them are same, could those steps be extracted and shared by 3 jobs so that just 1 place need to be changed?

@dsr20030703 Are you familiar with how to consolidate shared steps among GHA workflows like this? I'm not as familiar, but if you are, I would love a PR for that sort of thing. The duplication between the files bothers me too.

> Also, it seems that [this project is going to adopt uv](#254)? It would help a lot in setting up since [Manim also recommends using uv](https://docs.manim.community/en/stable/installation/uv.html). Shall I wait until #254 gets reviewed and merged?

I would say yes, probably. @eliotwrobson, do you have any thoughts on this?

eliotwrobson:
@caleb531 I agree, I think #254 is pretty close as well, so we should be able to close that out soon. I need a little bit of input from you there since I'm not 1000% how you want the deployment set up, and there's ways to consolidate the number of actions run if we would like.

dsr20030703:
> > It seems that `build.yml`, `lint.yml` and `tests.yml` shall be changed simultaneously when adding new dependencies. As the first 4 steps of them are same, could those steps be extracted and shared by 3 jobs so that just 1 place need to be changed?
> 
> @dsr20030703 Are you familiar with how to consolidate shared steps among GHA workflows like this? I'm not as familiar, but if you are, I would love a PR for that sort of thing. The duplication between the files bothers me too.

I’m not familiar with that either. 😢  Seems that we can combine these 3 files into 1 file, then extract the 4 common steps into a new job, like `setup`, then let the other 3 jobs [needs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) the `setup` job. So the combined yaml file would be like this:
```yaml
name: CI

…

jobs:
  setup:
    …

    steps:
      # the common 4 steps

  build:
    needs: setup
    …

  lint:
    needs: setup
    …

  tests:
    needs: setup
    …
```

eliotwrobson:
@dsr20030703 sorry for the delay on our end. Just merged the UV PR to the develop branch, so you should be good to rebase + update the code here.

dsr20030703:
@eliotwrobson I'm a newbie on git. How should I rebase and update code? 😥 I asked ChatGPT. The answer is like this:
```sh
git remote add caleb531 https://github.com/caleb531/automata.git
git fetch caleb531
git rebase caleb531/develop
# ... fix conflicts
git push origin animation --force
```
Is that OK? A senior of mine in school once told me: "there's only 1 occasion that you need to use `--force` in git -- You're gonna to be fired." 😰 Would it be better to use merge rather than rebase?

eliotwrobson:
The commands you posted are generally the correct way to do a rebase. You can take a look at this article as well: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase

> The problem is that the prior commits has been uploaded to "origin/animation" branch on GitHub. Before the last line of command, the local "animation" branch and the "origin/animation" branch on GitHub will separate. So the last line of command will discard the prior commits on "origin/animation" and replace them with new local commits. Is that OK?

If you're concerned about local / remote branches being out of sync, you should git pull before doing a rebase. If you do this first, I think it should be fine. But yes, a rebase does get rid of the old commits and replace them with the new ones, which is how the conflicts get resolved.

> A senior of mine in school once told me: "there's only 1 occasion that you need to use `--force` in git -- You're gonna to be fired." 😰

This is not correct, force pushes are a standard part of a rebase workflow, and many places strongly prefer this over merge commits to make it easier to do code review. This can be true if your organization doesn't have branch protection enabled, but we have this as part of this project. In short, by doing this you won't wipe out anybody else's work.

dsr20030703:
It seems that `manim` wasn't installed when [creating virtual environment in "Run mypy"](https://github.com/caleb531/automata/actions/runs/14888290879/job/41816024960#step:7:10)? 😕 I just ran command `uv add --optional visual manim` to add the requirements for "manim" to "visual" option (also add the steps in those 3 yaml files) in commit 34b3b74, then rebased to that commit. @eliotwrobson How did you made the `pyproject.toml` file? Are there any other configurations have to be made to let `manim` be installed during "Run mypy"?

eliotwrobson:
@dsr20030703 the error you're seeing is because the `manim` package is installed, but the library doesn't export the type annotations it has (I think, it looks like there isn't a py.typed file in their github: https://github.com/ManimCommunity/manim). Maybe you can import from the typing module explicitly? https://docs.manim.community/en/stable/contributing/docs/types.html

It might be worth adding an issue on their repo to get some clarity on what's going on. Worst case, you can add `manim` to the ignore list, which isn't a huge deal since the tests with this library are running + passing: https://github.com/caleb531/automata/blob/develop/pyproject.toml#L86-L94

dsr20030703:
I’ve opened an [issue](https://github.com/ManimCommunity/manim/issues/4247) there. While I’ve also found that there has been [an issue criticizing the code of manim (though feeling that it may not have a good title 😓) years ago](https://github.com/ManimCommunity/manim/issues/2380). As seems that it may take time to complete the issue, shall we add `manim` to the ignore list first?

[Another problem](https://github.com/caleb531/automata/actions/runs/14888290879/job/41816024960#step:7:24) I encountered in type checking is what discussed [here](https://github.com/caleb531/automata/pull/252#discussion_r2018135592). The return type of `partial` is `partial[M]` which conforms to `Callable[…, M]`. But in that case, one can’t get the argument list of the initializing function of `MObject` when coding, which would decrease the help of type annotation. Are there any better solutions? 🤔

eliotwrobson:
> As seems that it may take time to complete the issue, shall we add `manim` to the ignore list first?

I think that's the way to go. We can revisit this list later if `manim` ever figures things out.

> [Another problem](https://github.com/caleb531/automata/actions/runs/14888290879/job/41816024960#step:7:24) I encountered in type checking is what discussed [here](https://github.com/caleb531/automata/pull/252#discussion_r2018135592). The return type of `partial` is `partial[M]` which conforms to `Callable[…, M]`. But in that case, one can’t get the argument list of the initializing function of `MObject` when coding, which would decrease the help of type annotation. Are there any better solutions? 🤔

I see. I think it would be fine to use a type ignore statement or similar because the use of `partial` erases the type annotation. Otherwise, I think it's ok to change the return type to something like `Callable[..., M]` and return a function defined within the method you're seeing the type error in that does the same thing as the `partial` call with argument forwarding. Either way, I don't think that this function is going to be used too much outside of first party code, so I don't think it's a huge deal to have a potentially inaccurate annotation for this.

dsr20030703:
IMG_1628

> If you can get the tests passing, then I can do a full review 👍🏽

So we can begin the full review? By clicking “Ready for review” button?

automata/base/animation.py
```diff
@@ -0,0 +1,254 @@
...
+    shape : Dot | Circle | VGroup
+        - Dot for the nullnode (the start of the edge pointing the initial node)
+        - Circle for non-final state
+        - VGroup of a doublecircle for final state
+    label: Optional[Text]
+        The name of the state (N/A for the nullnode)
+    """
+
+    shape: manim.Dot | manim.Circle | manim.VGroup
+    label: Optional[manim.Text]
```
eliotwrobson, line 75:
```suggestion
    label: manim.Text | None
```

automata/base/animation.py
```diff
@@ -0,0 +1,254 @@
...
+        Parameters
+        ----------
+        node : `pygraphviz.Node`
+            the node which to based on to construct a FA state.
+
+            `node`'s label is its `attr['label']` if given, otherwise its
+            `name` property. Its `attr` contains:
+            - 'fontsize': float str,
+            - 'height': float str,
+            - 'pos': f'{float str},{float str}',
```
eliotwrobson, line 88: Why is there an f-string on this line?

automata/base/animation.py
```diff
@@ -0,0 +1,254 @@
...
+            - 'shape': 'point' or 'circle' or 'doublecircle',
+            - 'width': float str (possibly equals to 'height')
+        """
+        super().__init__(name=node.name)
+        radius = float(node.attr["height"]) / 2
+        if node.attr["shape"] == "point":
+            self.shape = Animate.default_init(manim.Dot)(radius=radius)
+            self.add(self.shape)
+        elif node.attr["shape"].endswith("circle"):
```
eliotwrobson, line 97: Is there no third case? Should we raise an exception if the shape does not meet either of these two conditions?

automata/base/animation.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Returns
+        -------
+        ApplyMethod
+            The animation for the `Scene` object to `play` to highlight the mobject.
+        """
+        return manim.FadeToColor(mobject, cls.HIGHLIGHT_COLOR)
+
+
+@dataclass(eq=False)
```
eliotwrobson, line 57: Why do you need to use the dataclass decorator if you have an init function?

automata/base/utils.py
```diff
@@ -26,12 +28,24 @@
 from frozendict import frozendict
 
 # Optional imports for use with visual functionality
-try:
+_missing_visual_imports: Optional[ImportError] = (
```
eliotwrobson, line 31: Same type annotation comment as above.

automata/fa/animation.py
```diff
@@ -0,0 +1,376 @@
...
+from automata.base.animation import Animate, _ManimEdge, _ManimInput, _ManimNode
+from automata.base.exceptions import RejectionException
+from automata.fa.fa import FA, FAStateT
+
+if TYPE_CHECKING:
+    from automata.fa.dfa import DFA, DFAStateT
+    from automata.fa.nfa import NFA, NFAStateT
+
+
+class _FAGraph(manim.VDict):
```
eliotwrobson, line 18: Is inheritance the right choice here? Or does the rest of the code only call the methods you define within this class? (i.e. the code here doesn't use methods inherited from `manim.VDict`).

automata/fa/animation.py
```diff
@@ -0,0 +1,376 @@
...
+        The DFA object.
+    input_str : str
+        The string to identify.
+    dfa_graph : _FAGraph
+        The graph of the DFA.
+    input_symbols : _ManimInput
+        The element in animation to show the current symbol of the `input_str`.
+    """
+
+    dfa: "DFA"
```
eliotwrobson, line 176: I think the string is not needed if you have `from __future__ import annotations`

pyproject.toml
```diff
@@ -41,7 +41,11 @@ classifiers = [
 
 # Per https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies
 [project.optional-dependencies]
-visual = ["coloraide>=1.8.2", "pygraphviz>=1.10"]
+visual = [
+    "coloraide>=1.8.2",
+    "manim>=0.19.0",
+ "pygraphviz>=1.10",
```
eliotwrobson, line 47: Could you make the indentation uniform?

eliotwrobson (approved):
A few small questions but overall the code looks really good! Two questions:

- Would it be possible to add some short discussion to the documentation? This can just point to the notebook since that's a more comprehensive example, but some mention of that would be good.
- Is there a way to double check that everything works without having `manim` installed? It would be good to make sure that we don't accidentally break anything because of the new optional dependency.

Want to check with @caleb531 to see what he thinks about the changes here but overall I think this has a significant value add for people using the library in jupyter notebooks.

automata/base/animation.py
dsr20030703: I just meant the value of `'pos'` attribute is a string of a decimal number, followed by a `,`, then followed by a string of another decimal number. I put an `f` here just meant that the text in the `{}` is a string of a decimal number, not a verbatim `float str`. The `f` can be ommited if it's confusing.

dsr20030703:
```suggestion
            - 'pos': '{float str},{float str}',
```

automata/fa/animation.py
dsr20030703: The only method used that's inherited from `manim.VDict` is the `[]` operator (i.e. `__getitem__`), by using which we get the nodes and edges of the graph. I tried to use a `manim.VDict` object to represents a graph, but later I found that I have to add new method. As a graph can be represented by a dictionary, I think that graph can *be* a dictionary, thus I use inheritance.

pyproject.toml
dsr20030703: > I just ran command `uv add --optional visual manim` to add the requirements for "manim" to "visual" option (also add the steps in those 3 yaml files) in commit (34b3b74), then rebased to that commit. 

`uv` did that, and I prefer to leave it. In my opinion, as we've adopted `uv` as project manager, we'd better modify these files by ourselves as least as possible, and use `uv` command to manage these files.

eliotwrobson: We should be formatting these files (possibly using an autoformatter), as uv does not do this. uv also does not manage the project metadata, which we must write / update manually, and consistent formatting helps make this more readable.

automata/fa/animation.py
eliotwrobson: How frequently is this operator used? Although a graph can be represented with a dictionary, we should avoid inheritance if possible and use composition (i.e. have the `VDict` as a member variable), since that exposes fewer methods to the client code. It would be better to have a couple limited methods instead of using the `[]` operator directly if this is only used in limited ways.

automata/base/animation.py
eliotwrobson: You do not need the dataclass decorator to do this: https://stackoverflow.com/questions/53897054/python-class-members-type-hinting

The dataclass decorator is used to avoid code duplication by automatically generating some methods based on member variables. Unless you are using those generated methods, then you should not be using this decorator.

automata/base/animation.py
eliotwrobson: This is true, although the function can theoretically take in an argument of any type, even if by mistake. I'd prefer to have an exception raised if neither case is hit to avoid a weird noop if the shape isn't recognized.

automata/base/animation.py
eliotwrobson: That makes sense. I think omitting the f is the way to go throughout this PR. I initially wasn't sure whether this was an f-string or not.

automata/fa/animation.py
dsr20030703: > How frequently is this operator used? 

Very Frequently. 😥 There’re lots of occasions where I need to get a node or an edge.

> we should avoid inheritance if possible and use composition (i.e. have the `VDict` as a member variable),

```suggestion
class _FAGraph:
    nodes: manim.VDict
    edges: manim.VDict
```

Would modification like that be OK? In that case, we would use `fa_graph.nodes[node_name]` to get a node and `fa_graph.edges[tail_name, head_name]` to get an edge. Or shall we add a `get_node` method and a `get_edge` method and let the `nodes` and `edges` private? (Though the whole class is already private)

eliotwrobson: I think the change as proposed is also ok. The prior inheritance makes more sense at least since the classes used are smaller, but here I think it's significantly simpler to avoid.

dsr20030703:
> * Would it be possible to add some short discussion to the documentation? This can just point to the notebook since that's a more comprehensive example, but some mention of that would be good.

There have been links to the `example_notebooks` folder in `docs/index.md` and `README.md`. Or perhaps we can rewrite the `docs/examples/fa-examples.md` according to the current 2 Jupyter Notebooks? After all, Jupyter Notebooks can't run directly on a webpage, but videos can. And we can use HTML `<video>` tag to embed videos in Markdown. 

> * Is there a way to double check that everything works without having `manim` installed? It would be good to make sure that we don't accidentally break anything because of the new optional dependency.

Currently, the public functions (i.e. whose not starting with a `_` and the name of the class it belongs, if there is, also doesn't start with a `_`) that needs `manim`, are only the `animate_reading_input` function in `DFA` and `NFA` class. And I've added the checks for `manim` in these 2 functions.

eliotwrobson:
> There have been links to the `example_notebooks` folder in `docs/index.md` and `README.md`. 

Ahh I forgot that this linked directly to the folder. I think it's fine for now then.

> Currently, the public functions (i.e. whose not starting with a `_` and the name of the class it belongs, if there is, also doesn't start with a `_`) that needs `manim`, are only the `animate_reading_input` function in `DFA` and `NFA` class. And I've added the checks for `manim` in these 2 functions.

That sounds good I think, we don't have to write an automated test and it seems like the impact is pretty minimal.

eliotwrobson:
@dsr20030703 did you get a chance to wrap up the remaining changes? If so, I'd like to review + merge this soon so things don't get too stale

dsr20030703:
May I ask what does “the remaining changes” refers to? If it refers to other changes not pushed yet, there’re none.

eliotwrobson:
@dsr20030703 sorry, I misread the prior discussion, it looks like all issues have been resolved. I'll do one more pass over this and I think we can merge soon (cc @caleb531 if you have any thoughts).

automata/base/animation.py
```diff
@@ -0,0 +1,256 @@
...
+            a str of spline pattern:
+            - spline = endp point (triple)+
+            - point = "%f,%f"
+            - endp = "e,"point
+            - triple = point point point
+        """
+        points = edge_pos.split()
+        control_points = tuple(
+            (*(float(pt) / _POINTS_IN_INCH for pt in point.split(",")), 0)
+            for point in points[1:]
```
eliotwrobson, line 182:
```suggestion
            for point in itertools.islice(points, 1, None)
```
Nit: Avoid making a copy.

automata/fa/animation.py
```diff
@@ -0,0 +1,380 @@
...
+                    current_transitions, lambda_transitions
+                ),
+                *self.nfa_graph.highlight_states(
+                    map(itemgetter(1), lambda_transitions)
+                ),
+            )
+            current_transitions = lambda_transitions
+        self.wait()
+        input_index = -1
+        for input_index in range(len(self.input_str)):
```
eliotwrobson, line 339:
```suggestion
        for input_index, input_symbol in enumerate(self.input_str):
```
Avoid the lookup of the input symbol by the index in every iteration. Also, delete the line above (no need to define this variable before the for loop).

eliotwrobson (approved):
@caleb531 any more thoughts on this PR? Have a couple of nitpick items but I think this is ready to go, and would be nice to have on the latest release.

caleb531:
@eliotwrobson @dsr20030703 Where did we end up with the conversation about whether `automata` can be installed without `manim` or some global library installed as a prerequisite? I was reading through the comment history, and it wasn't entirely clear. I just want to make sure that people who don't need the animation functionality don't need to install anything more to install `automata`. In other words, `manim` needs to be truly optional.

I think that is my only real question/concern at this point. Pending @eliotwrobson's suggestions and feedback, everything else looks great.

@eliotwrobson Lastly, shall we open a GitHub issue to research consolidating the shared steps of the GHA workflows? Some quick research and assistance from ChatGPT shows that this might help:
https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action

PR #254, opened 2025-04-02
Setting up UV

### LINKED ISSUE #253: Clean up install
Found by @nwalters512 (context omitted)

Image

@caleb531 we should remove the `tests` directory from the published version of the package. Might it be a good idea to use a more modern build tool like `uv` for publishing / dependency management? I used it for [another project](https://github.com/eliotwrobson/depqdict), and it was very straightforward to set up automated testing / publishing.

eliotwrobson:
Resolves #253


eliotwrobson:
@caleb531 so it appears this was a lot easier than expected. I don't think this is quite in a stage to merge yet (I'm not sure exactly how you want to set up the publish workflow), but I think this is a strong proof of concept so far. I might need your input on how we want to test this fully before merging, since this is a pretty sizeable change, but overall I think simplifies the CI configuration.

eliotwrobson:
@caleb531 do you have any thoughts about the publish workflow? That's the only one that I haven't yet changed in this PR. The docs building one should work but I haven't tested it.

caleb531:
@eliotwrobson I think you just need to update the publish workflow to build the project properly with `uv` to generate the wheel and other distributions (can't remember everything that gets generated). The step is called "Build a binary wheel and a source tarball" in publish.yml. But I think once the project is built in the context of the GHA workflow runtime, it gets deployed from there. Granted, I haven't found any pypa/gh-action-pypi-publish documentation related to uv.

If we want to test this, we can try pushing a dev or alpha prerelease or something. The publish workflow gets triggered when a tag is created, regardless of which branch it is on.

.github/workflows/publish.yml
```diff
@@ -18,41 +18,18 @@ jobs:
...
-          # disabling attestations on the test PyPI step should allow the
-          # production PyPI step to succeed
-          attestations: false
-
-      - name: Publish distribution to PyPI
-        uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          repository_url: https://upload.pypi.org/legacy/
+          uv build
+          uv publish
```
eliotwrobson, line 35: This should just work according to the docs here: https://docs.astral.sh/uv/guides/package/#publishing-your-package

eliotwrobson: @caleb531 do you think we should merge to develop and then debug on the fly during the next deployment? Then some of the other PRs can get merged in the meantime.

caleb531: @eliotwrobson I think that's fine. I'm reading that `uv publish` is compatible with the Trusted Publisher functionality in PyPI, so I feel optimistic about switching to `uv` for publishing. But we won't really know until we can test a publish from `develop`.

eliotwrobson: That makes sense. I'll go ahead and mark this as ready for review.

eliotwrobson:
@caleb531 did you have a chance to take a look at this? If not, I think it's probably ok to just merge and debug later.

calebevans-modeln:
@eliotwrobson Ah, my apologies. It's been a busy few weeks. Let me merge now.

calebevans-modeln (approved):

calebevans-modeln:
@eliotwrobson Strangely, it's not letting me merge

Screenshot 2025-05-06 at 2 12 01 PM

caleb531 (approved):

PR #264, opened 2025-10-31
Skip relevant tests if optional dependencies are missing

caleb531:
@eliotwrobson I was dialoguing with GitHub Copilot / GPT-5 today trying to troubleshoot an installation error with `manim` on my local machine. I ended up getting it resolved—the issue was mostly due to version mismatch between the project pygraphviz and my system graphviz installations.

Interestingly, Copilot suggested that, in order to reduce impact of installation errors like this, to update the tests so that optional dependencies are imported lazily in a way that skips the visual tests if optional dependencies are not installed. So I had it make these changes to see how it would look, and the result is this PR.

I'm a bit on the fence about whether this is the right approach to take.
1. On one hand, it could be argued that _all_ of our tests should run every time.
2. But then again, for contributors who don't need to interact with the visualization features, the failing visual tests (due to system installation errors) would probably get in the way of their local development.

What do you think?

### DIFF
--- a/tests/optional.py
+++ b/tests/optional.py
@@ -1,0 +1,32 @@
+"""Test helpers for optional dependencies.
+
+Provides a VISUAL_OK flag indicating whether visualization extras are
+importable (pygraphviz, coloraide). Use in tests to conditionally skip
+diagram-related tests on environments lacking these extras.
+"""
+
+from __future__ import annotations
+
+# Shared skip reason for tests that require visualization extras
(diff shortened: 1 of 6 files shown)


eliotwrobson (approved):
I think this makes sense, I've seen other packages do similar

caleb531:
@eliotwrobson Ah good to know we'd be following suit with other packages. Alright, will merge this then!

PR #266, opened 2025-11-03
Switch from nose2 to pytest for all tests

caleb531:
@eliotwrobson Here is my PR to switch from nose2 to pytest. I explored a few different approaches, including dissolving the classes and switching to `assert *` syntax. However, this provided to be too significant of an undertaking, so I kept it simple by preserving the `unittest`-based classes.

The only change I needed to make to the testing code was for parameterization (only used in `tests.test_dfa`). After some experimentation and research, I found that pytest's `@pytest.mark.parametrize` decorator does not support `unittest.TestCase` methods because of the way `@pytest.mark.parametrize` handles parameter injection (which is very different from `unittest`). But I was able to pull in the [`parameterized`](https://pypi.org/project/parameterized/) package as an additional dev dependency, which made the switchover pretty seamless.

@dsr20030703 Please let me know if these changes would enable you to integrate the Graphical Unit Tests and Video Format Tests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,9 +41,9 @@ In addition, new code must include type annotations and pass typechecking run wi
 2. Whether you are introducing a bug fix or a new feature, you *must* add tests
 to verify that your code additions function correctly and break nothing else.
 
-1. Please run `uv run pytest --cov` and ensure that your changes are covered.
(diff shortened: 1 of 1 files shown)


caleb531:
@eliotwrobson The force pushes were just to fix a few typos that were breaking CI, but all checks are passing now. This PR is indeed ready for your review. 👍

PR #286, opened 2026-02-07
Migrate from mypy to ty

caleb531:
So the creators of uv recently promoted their type-checker, [ty](https://docs.astral.sh/ty/) to be production-ready, and after being pleased with its adoption for my other Python projects, I thought it would be a better type checker to have here, as well.

This PR updates the relevant dependencies and configuration to replace mypy with ty, removing old mypy references (except the gitignore because that can create unchecked files for people who haven't yet migrated). I've also updated the source code and tests to strike a balance between type strictness (since ty is stricter) and type relaxation (e.g. for tests).

Let me know what you think! GPT-5.2-Codex wrote most of the code, but I supervised its operations and took extra steps to fix type errors rather than simply disabling rules on a per-file basis.


eliotwrobson (approved):
Just a spot check but overall looks like a big improvement! No big changes and removed a lot of ignore comments.

caleb531:
@eliotwrobson Great great! In that case, I will proceed to merge into `develop`.

